PR #3, opened 2019-09-30
fix: disables runtime tests on all platform but windows

baszalmstra:
A better fix would be to fix the tests but this takes a lot more work and currently the travis build fails. We'll have to revisit this later.


baszalmstra:
[Disable runtime tests until they are fixed](https://trello.com/c/QJykCIdg/40-disable-runtime-tests-until-they-are-fixed)

PR #5, opened 2019-10-03
feat: adds detection of duplicate definition names

baszalmstra:
The thing that is still missing in the PR is a test to actually check that this works but currently there is no testing framework to test HIR in the first place.


baszalmstra:
[Error on duplicate symbols](https://trello.com/c/CZ0oOTBB/17-error-on-duplicate-symbols)

PR #10, opened 2019-10-05
feat: use azure pipelines instead of travis

baszalmstra:
For an example of the result see:
https://dev.azure.com/mun-lang/mun/_build/results?buildId=49&view=results

Currently the pipeline checks the code with `cargo check`, runs `cargo test` for stable, beta, nightly on Linux, and computes codecoverage.

Both the interface is nicer, and its builds considerably faster than travis.

- [x] Builds & tests on Linux
- [x] Builds & tests on Windows
- [x] Builds & tests on macOS
- [x] Builds Beta&Nightly
- [x] Checks formatting with `cargo fmt --check`
- [x] Uploads code coverage to codecov
- [x] Update readme badges


baszalmstra:
[Checkout out azure pipelines to support all target platforms](https://trello.com/c/zZZEAmgj/43-checkout-out-azure-pipelines-to-support-all-target-platforms)

PR #12, opened 2019-10-11
Unit tests for auto-generated ABI structs

Wodann:
Depends on PR #11


Wodann:
There's actually a difference between _fakes_ and _mocks_ (and _stubs_):

- https://stackoverflow.com/questions/346372/whats-the-difference-between-faking-mocking-and-stubbing
- https://blog.pragmatists.com/test-doubles-fakes-mocks-and-stubs-1a7491dfa3da

PR #15, opened 2019-10-18
Add an extension for result objects that allows retrying on failure

Wodann:


crates/mun_runtime/src/lib.rs
```diff
@@ -155,27 +156,37 @@ impl MunRuntime {
     }
 }
 
+/// Extends a result object with functions that allow retrying of an action.
+pub trait Retriable: Sized {
```
Wodann: Are we happy with this trait name? Or would something like `RetryResultExt` or `RetryExt` be better? `Ext` meaning extension, as this is a trait extension.

baszalmstra: RetryResultExt sounds good

PR #18, opened 2019-10-21
Solve clippy warnings

Wodann:
Resolved most clippy warnings


Wodann:
@baszalmstra it does not. There are still some clippy errors left due to salsa attribute macros that I wasn't able to solve easily.

PR #29, opened 2019-10-30
feat: adds update operators and diagnostics

baszalmstra:
Adds `+=`, `-=`, `*=`, `/=` operators.

A lot of files seem to have changed but this is mostly due to how IR is generated for functions parameters. Now that all locals including parameters are mutable (mutability is for another PR) we need to be able to get a pointer to the storage of a parameter. This is not possible without allocating the locals on the stack first. This already happens for locally introduced variables but not for function parameters. For parameters an `alloca` is first performed, then the function parameter value is stored on the stack. LLVM optimizes this out again so there is no overhead. However, the tests differ because now the stack based parameter has the name of the parameter instead of the function parameter. This results in slightly differently generated IR naming but the operations are still the same.


crates/mun_hir/src/ty/tests.rs
```diff
@@ -51,6 +51,18 @@ fn void_return() {
     )
 }
 
+#[test]
+fn place_expressions() {
+    infer_snapshot(
+        r#"
+    fn foo(a:int) {
+        a += 3;
```
Wodann: To be "feature-complete" it would be good to also add `-=`, `*=`, and `/=` to the test.

crates/mun_codegen/src/test.rs
```diff
@@ -106,6 +106,18 @@ fn update_operators() {
     );
 }
 
+#[test]
+fn update_parameter() {
+    test_snapshot(
+        r#"
+    fn add_three(a:int):int {
+      a += 3;
```
Wodann: To be "feature-complete" it would be good to also add `-=`, `*=`, and `/=` to the test.

crates/mun_hir/src/ty/tests.rs
baszalmstra: Ill add another test specifically for that

crates/mun_codegen/src/test.rs
baszalmstra: Ill add another test specifically for that

baszalmstra: Thats not the purpose of this test, this test just checks if updating a function parameter works in the first place. Otherwise Id have to add `+`, `-` etc too.

baszalmstra:
[Enable mutation of variables](https://trello.com/c/Mys6rfun/73-enable-mutation-of-variables)

PR #34, opened 2019-11-03
fix(compiler): allow absolute file paths

Wodann:
This assumes that the provided file is in the root directory.


Wodann:
[Cannot use absolute paths with the mun_compiler](https://trello.com/c/GEyBI7ot/79-cannot-use-absolute-paths-with-the-muncompiler)

PR #42, opened 2019-11-07
test(cli): adds test to check gen-syntax ran

baszalmstra:


crates/tools/src/lib.rs
```diff
@@ -1,25 +1,84 @@
...
+    let content = teraron::render(&template, src)?;
+    let content = reformat(content)?;
+    update(&tgt, &content, mode)
+}
+
+/// A helper to update file on disk if it has changed.
+/// With verify = false,
+fn update(path: &Path, contents: &str, mode: Mode) -> Result<()> {
+    match fs::read_to_string(path) {
+        Ok(ref old_contents) if old_contents == contents => {
```
Wodann: My only concern is how expensive it might be.

PR #43, opened 2019-11-08
test(hir): checks if source computations are properly cached

baszalmstra:


baszalmstra:
[Write test to make sure module definitions dont change on reparse](https://trello.com/c/t4juhUMb/50-write-test-to-make-sure-module-definitions-dont-change-on-reparse)

PR #46, opened 2019-11-10
fix: fix boolean literal conversion

baszalmstra:
This PR fixes an issue with boolean literals in the conversion from ast to HIR. I've also added two tests to test the functionality.


Wodann:
[The Mun bool type does not support interop with Rust](https://trello.com/c/gBh7Vryb/96-the-mun-bool-type-does-not-support-interop-with-rust)

Wodann:
[Add regression tests for data interop](https://trello.com/c/jOqW40FC/97-add-regression-tests-for-data-interop)

PR #49, opened 2019-11-15
Incremental compilation & hotreloading tests

baszalmstra:
This commit restructures the compiler crates to better support incremental compilation. The compiler `--watch` mode is now incremental, reusing the state from previous compilations.

**Todo:**

- [x] Implement `Driver`
- [x] Make `--watch` mode incremental with `Driver`
- [x] Add a test to check `--watch` mode works
- [x] Add a runtime test to verify hotloading

### DIFF
--- a/crates/mun_compiler/src/diagnostics/emit.rs
+++ b/crates/mun_compiler/src/diagnostics/emit.rs
@@ -42,7 +42,7 @@ impl Emit for Diagnostic {
         writer.set_color(&error)?;
         write!(writer, "error")?;
 
-        // Write diagnostics message
+        // Write diagnostic message
         writer.set_color(&header)?;
         writeln!(writer, ": {}", self.message)?;
 
@@ -74,7 +74,7 @@ impl Emit for Diagnostic {
             writer.set_color(&error)?;
 
             if line_col.line == line_col_end.line {
-                // single-line diagnostics
+                // single-line diagnostic
                 writeln!(
                     writer,
                     " {}{}",
--- a/crates/mun_compiler/src/diagnostics/emit.rs
+++ b/crates/mun_compiler/src/diagnostics/emit.rs
@@ -42,7 +42,7 @@ impl Emit for Diagnostic {
         writer.set_color(&error)?;
         write!(writer, "error")?;
 
-        // Write diagnostic message
+        // Write diagnostics message
         writer.set_color(&header)?;
         writeln!(writer, ": {}", self.message)?;
 
@@ -74,7 +74,7 @@ impl Emit for Diagnostic {
             writer.set_color(&error)?;
 
             if line_col.line == line_col_end.line {
-                // single-line diagnostic
+                // single-line diagnostics
                 writeln!(
                     writer,
                     " {}{}",
--- a/crates/mun_compiler/src/driver.rs
+++ b/crates/mun_compiler/src/driver.rs
@@ -29,7 +29,7 @@ pub struct Driver {
 }
 
 impl Driver {
-    /// Construct a driver with a specific configuration.
+    /// Constructs a driver with a specific configuration.
     pub fn with_config(config: Config) -> Self {
         let mut driver = Driver {
             db: CompilerDatabase::new(),
@@ -106,7 +106,7 @@ impl Driver {
             .collect()
     }
 
-    /// Emits all diagnostic messages currently in the database; returns true if errors where
+    /// Emits all diagnostic messages currently in the database; returns true if errors were
     /// emitted.
     pub fn emit_diagnostics(&self, writer: &mut impl WriteColor) -> Result<bool, failure::Error> {
         let mut has_errors = false;
--- a/crates/mun_compiler/src/driver.rs
+++ b/crates/mun_compiler/src/driver.rs
@@ -48,7 +48,7 @@ impl Driver {
         driver
     }
 
-    /// Constructs a driver with a configuration
+    /// Constructs a driver with a configuration and a single file.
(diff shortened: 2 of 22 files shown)


baszalmstra:
[Add incremental compilation to compiler daemon by properly using salsa](https://trello.com/c/pVl2KU35/51-add-incremental-compilation-to-compiler-daemon-by-properly-using-salsa)

baszalmstra:
[Create tests for hot reloading functionality](https://trello.com/c/4r3xDxWh/59-create-tests-for-hot-reloading-functionality)

crates/mun_compiler/src/diagnostics/emit.rs
```diff
@@ -42,7 +42,7 @@ impl Emit for Diagnostic {
         writer.set_color(&error)?;
         write!(writer, "error")?;
 
-        // Write diagnostic message
+        // Write diagnostics message
```
Wodann: singular is diagnostic, plural is diagnostics.

crates/mun_compiler/src/driver.rs
```diff
@@ -0,0 +1,157 @@
...
+pub const WORKSPACE: SourceRootId = SourceRootId(0);
+
+#[derive(Debug)]
+pub struct Driver {
+    db: CompilerDatabase,
+    out_dir: Option<PathBuf>,
+}
+
+impl Driver {
+    /// Construct a driver with a specific configuration.
```
Wodann:
```suggestion
    /// Constructs a driver with a specific configuration.
```

crates/mun_compiler/src/driver.rs
```diff
@@ -0,0 +1,157 @@
...
+    pub fn with_config(config: Config) -> Self {
+        let mut driver = Driver {
+            db: CompilerDatabase::new(),
+            out_dir: None,
+        };
+
+        // Move relevant configuration into the database
+        driver.db.set_target(config.target);
```
Wodann: Would a single `set_config` be an option? That would simplify adding options and reduce *failure points*.

crates/mun_compiler/src/driver.rs
```diff
@@ -0,0 +1,157 @@
...
+    pub fn diagnostics(&self) -> Vec<Diagnostic> {
+        self.db
+            .source_root(WORKSPACE)
+            .files()
+            .map(|f| diagnostics(&self.db, f))
+            .flatten()
+            .collect()
+    }
+
+    /// Emits all diagnostic messages currently in the database; returns true if errors where
```
Wodann:
```suggestion
    /// Emits all diagnostic messages currently in the database; returns true if errors were
```

crates/mun_runtime/src/test.rs
```diff
@@ -1,95 +1,143 @@
...
+        TestDriver {
+            _temp_dir: temp_dir,
+            driver,
+            out_path,
+            file_id,
+            runtime,
+        }
+    }
+
+    /// Updates the text of the Mun source and ensure the generated assembly has been reloaded.
```
Wodann:
```suggestion
    /// Updates the text of the Mun source and ensures that the generated assembly has been reloaded.
```

crates/mun_runtime/src/test.rs
```diff
@@ -1,95 +1,143 @@
...
+            if now - start_time > std::time::Duration::from_secs(10) {
+                panic!("runtime did not update after recompilation within 10secs");
+            } else {
+                sleep(Duration::from_millis(1));
+            }
+        }
+    }
+
+    /// Calls the function with the specified name
+    fn invoke0<T: Reflection>(&mut self, fn_name: &str) -> T {
```
Wodann: Why don't you add a `runtime_mut()` function to `TestDriver`? That way you don't have to duplicate the `invoke_fn` logic.

crates/mun_compiler/src/diagnostics/emit.rs
baszalmstra: I think I renamed a field which also renamed comments ...

crates/mun_compiler/src/driver.rs
baszalmstra: Do you mean on the `Driver` or on the `CompilerDatabase`? There are two considerations

1. The `Config` can contain options for both and/or either the `Driver` and the `Database`.
2. Constructing a `Driver` or a `CompilerDatabase` without configuration doesnt make sense.

crates/mun_runtime/src/lib.rs
```diff
@@ -131,7 +131,7 @@ impl Runtime {
         assembly.link(&self.dispatch_table)?;
 
         self.watcher
-            .watch(library_path.clone(), RecursiveMode::NonRecursive)?;
+            .watch(library_path.parent().unwrap(), RecursiveMode::NonRecursive)?;
```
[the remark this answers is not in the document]
baszalmstra: The first line of the method canonicalizes the path so it should be an existing absolute path.

crates/mun_runtime/src/test.rs
baszalmstra: Because this is shorter which I find much clearer.

crates/mun_compiler/src/driver.rs
Wodann: The compiler database. Now you need to manually set all of the different config fields.

crates/mun_runtime/src/test.rs
Wodann: The trade-off is that a lot more code is added overall.

PR #62, opened 2019-11-23
Adds `break` expressions

baszalmstra:
This PR adds the `break` expression to Mun. It allows breaking from loops and enables loops to have return types different from `never`.

The following code should work properly.

```mun
fn fibonacci(n:int):int {
    let a = 0;
    let b = 1;
    let i = 1;
    loop {
        if i > n {
            break a;
        }
        let sum = a + b;
        a = b;
        b = sum;
        i += 1;
    }
}
```

Here the `loop` should get the `int` type instead of the `never` type due to the `break a` statement.

- [x] Implement parsing of break expressions.
- [x] Implement proper type inferencing and diagnostics
  - [x] Do not allow break expression outside loops
  - [x] Multiple break expressions must have the same type 
- [x] Implement code generation for the break statement

**Depends on #60**


baszalmstra:
[Break expression](https://trello.com/c/JfzaMlLP/99-break-expression)

crates/mun_syntax/src/tests/parser.rs
```diff
@@ -186,3 +186,16 @@ fn loop_expr() {
...
     )
 }
+
+#[test]
+fn break_expr() {
+    ok_snapshot_test(
+        r#"
+    fn foo() {
+        break;
+        if break { 3; }
```
Wodann: Is this supposed to work? What is the meaning?

baszalmstra: Yeah, it works! Its a bit strange but also works with Rust.

PR #63, opened 2019-11-23
Adds `while` expressions

baszalmstra:
This PR adds `while` expressions to Mun. 

The following code should work properly.

```
fn fibonacci(n:int):int {
    let a = 0;
    let b = 1;
    let i = 1;
    while i <= n {
        let sum = a + b;
        a = b;
        b = sum;
        i += 1;
    };
    a
}
```

- [x] Implement parsing of while expressions.
- [x] Implement proper type inferencing and diagnostics
  - [x] Do not allow break expression with values inside while
- [x] Implement code generation for the while statement

**Depends on #62**


baszalmstra:
[While statements](https://trello.com/c/vCbrmBfu/64-while-statements)

crates/mun_codegen/src/ir/body.rs
```diff
@@ -69,6 +70,7 @@ impl<'a, 'b, D: IrDatabase> BodyIrGenerator<'a, 'b, D> {
             function_map,
             dispatch_table,
             active_loop: None,
+            hir_function,
```
Wodann: What function is this? The parent function that this body is part of?

crates/mun_hir/src/ty/tests.rs
```diff
@@ -144,6 +144,21 @@ fn infer_break() {
...
 
+#[test]
+fn infer_while() {
+    infer_snapshot(
+        r#"
+    fn foo() {
+        let n = 0;
+        while n < 3 { n += 1; };
+        while n < 3 { n += 1; break; };
+        while n < 3 { break 3; };   // Error: break with value in while
```
Wodann: Should this match the actual error?
`break` with value can only appear in a `loop`

crates/mun_codegen/src/ir/body.rs
baszalmstra: This is the HIR representation of the function that the body is part of. Its required to determine the return type.

Wodann: Okay. I was wondering whether there was a more verbose way of naming `hir_function`, but cannot quickly come up with anything.

PR #67, opened 2019-12-22
test: add test to check if compiler output strings is valid utf8

### LINKED ISSUE #52: Test whether mun_compiler generated strings are valid UTF-8
- [x] claim this issue (assign yourself or comment below)
- [ ] [setup repository on your local machine](https://github.com/mun-lang/mun) and make sure all tests pass (`cargo test`)
- [ ] read our [contributing guidelines](https://github.com/mun-lang/mun/blob/master/CONTRIBUTING.md)
- [ ] the Mun Compiler [generates UTF-8 strings](https://github.com/mun-lang/mun/blob/master/crates/mun_codegen/src/code_gen/symbols.rs) in the Mun ABI. For every such function, add a test that check that the resulting shared library contains valid UTF-8.
- [ ] start a Pull Request. Set description to `closes #52`. If this is your first PR, welcome :tada: :smile:

legendiguess:
closes #52

### DIFF
--- a/crates/mun_runtime/src/test.rs
+++ b/crates/mun_runtime/src/test.rs
@@ -298,3 +298,39 @@ fn hotreloadable() {
     );
     assert_invoke_eq!(i64, 10, driver, "main");
 }
+
+#[test]
+fn compiler_valid_utf8() {
+    use std::ffi::CStr;
+    use std::str;
+
+    let mut driver = TestDriver::new(
+        r#"
+    fn foo(n:int):bool { false }
+    "#,
+    );
+
+    let foo_func = driver.runtime.get_function_info("foo").unwrap();
+
+    // Check foo name for valid utf8
+    assert_eq!(
+        str::from_utf8(&unsafe { CStr::from_ptr(foo_func.signature.name).to_bytes() }).is_err(),
+        false
+    );
+    // Check foo arg type for valid utf8
+    assert_eq!(
+        str::from_utf8(&unsafe { CStr::from_ptr((*foo_func.signature.arg_types).name).to_bytes() })
+            .is_err(),
+        false
+    );
+    // Check foo return type for valid utf8
+    assert_eq!(
+        str::from_utf8(&unsafe {
+            CStr::from_ptr((*foo_func.signature.return_type).name).to_bytes()
+        })
+        .is_err(),
+        false
+    );
+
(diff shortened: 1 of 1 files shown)


crates/mun_runtime/src/test.rs
```diff
@@ -298,3 +298,55 @@ fn hotreloadable() {
...
+    // Check foo return type for valid utf8
+    assert_eq!(
+        str::from_utf8(&unsafe {
+            CStr::from_ptr((*foo_func.signature.return_type).name).to_bytes()
+        })
+        .is_err(),
+        false
+    );
+
+    // Check main name for valid utf8
```
Wodann: In essence, this and the following tests are the same as the tests for `foo_func`s tests, so they are unnecessary. If the logic is valid for one test case (that runs through all code paths) another test case will give the same results.

Since Mun doesn't require a main function, you can actually remove the `main` function from the example and remove the corresponding assertions.

crates/mun_runtime/src/test.rs
```diff
@@ -298,3 +298,55 @@ fn hotreloadable() {
     );
     assert_invoke_eq!(i64, 10, driver, "main");
 }
+
+#[test]
+fn compiler_vaild_ut8() {
```
mswf:
```suggestion
fn compiler_valid_utf8() {
```

crates/mun_runtime/src/test.rs
legendiguess: Good remark!

crates/mun_runtime/src/test.rs
legendiguess: Thanks! Silly typo :see_no_evil:

crates/mun_runtime/src/test.rs
```diff
@@ -298,3 +298,39 @@ fn hotreloadable() {
     );
     assert_invoke_eq!(i64, 10, driver, "main");
 }
+
+#[test]
+fn compiler_vaild_utf8() {
```
Wodann:
```suggestion
fn compiler_valid_utf8() {
```

legendiguess: i fix it right now!

baszalmstra:
[Create tests that validate assumptions about interop between compiler and runtime](https://trello.com/c/IqE30qjs/58-create-tests-that-validate-assumptions-about-interop-between-compiler-and-runtime)

PR #69, opened 2019-12-29
feat: add manual generation of abi, runtime-capi

baszalmstra:
When using build.rs files several files get regenerated every build
which causes rebuilds even when nothing really changed. This commit
changes this behavior to make the user manually run a command when the
files should be regenerated.


Wodann:
Is it necessary to do this for the CAPI too?

baszalmstra:
> Is it necessary to do this for the CAPI too?

Appears to be the case yes. Without this cargo always detects changes.

.github/workflows/ci.yml
```diff
@@ -61,7 +61,9 @@ jobs:
 
       - name: Cargo check
         uses: actions-rs/cargo@v1
-        continue-on-error: ${{ matrix.config.toolchain != 'stable' }}
+        env:
```
Wodann: Are you reverting this from your previous change?

crates/tools/src/abi.rs
```diff
@@ -1,5 +1,9 @@
-use std::env;
-use std::path::PathBuf;
+use crate::{project_root, reformat, update, Result};
+use failure::format_err;
+use teraron::Mode;
+
+pub const RUNTIME_CAPI_DIR: &str = "crates/mun_runtime_capi";
```
Wodann:
```suggestion
```

This constant is not used in this file.

.github/workflows/ci.yml
baszalmstra: Yes, I'm fixing this because the code should just build with Beta IMO

Wodann: Okay. Maybe split this into a separate commit to clarify this (like you did for the previous PR)?

baszalmstra: 👍 reverted

Wodann:
Do you have any idea why the gen-runtime-capi check is failing just for ubuntu beta?

baszalmstra:
> Do you have any idea why the gen-runtime-capi check is failing just for ubuntu beta?

No, I cant reproduce it either. But Im working on it. :)

baszalmstra:
[Remove `runtime-ffi` submodule from `mun` repository](https://trello.com/c/0GtG3Vhn/98-remove-runtime-ffi-submodule-from-mun-repository)

PR #81, opened 2020-01-13
fix: struct initialization fix for non constants

baszalmstra:
Fixes an issue when initializing structs with non constant values resulted in a crash.

### DIFF
--- a/crates/mun_codegen/src/ir/body.rs
+++ b/crates/mun_codegen/src/ir/body.rs
@@ -11,9 +11,8 @@ use inkwell::{
 };
 use std::{collections::HashMap, mem, sync::Arc};
 
-use crate::ir::adt::gen_named_struct_lit;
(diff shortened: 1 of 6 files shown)


crates/mun_codegen/src/ir/body.rs
```diff
@@ -218,9 +217,19 @@ impl<'a, 'b, D: IrDatabase> BodyIrGenerator<'a, 'b, D> {
     fn gen_struct_alloc(
         &mut self,
         hir_struct: hir::Struct,
-        args: &[BasicValueEnum],
+        args: Vec<BasicValueEnum>,
```
Wodann: Why not a slice? That'd more universal.

baszalmstra: Because I want to consume the contents completely.

PR #82, opened 2020-01-14
fix: link against ucrt.lib to resolve missing symbols for malloc

Wodann:
Temporary fix for Windows, as long as we are using `malloc`.


Wodann:
No longer necessary as PR #90 introduces _runtime linked_ `malloc`.

PR #83, opened 2020-01-14
Add marshalling of structs

Wodann:
Depends on #90


Wodann:
That's totally fine. The temporary fix is cumbersome in CI, so better not waste any more time on it

Wodann:
[Add marshalling of structs to Rust](https://trello.com/c/jEOrWCyF/116-add-marshalling-of-structs-to-rust)

PR #84, opened 2020-01-20
Refactor type information to dynamically include struct information

Wodann:
Depends on #83.

`TypeInfo` is now a globally allocated type that can potentially have additional information appended, depending on its `TypeGroup`. As a result, all places that passed `MunTypeInfo` by value now use a `MunTypeInfo*`

This is not reflected in the `mun_abi.h`, as including a `union` type would affect the size of all `MunTypeInfo` - not just the ones that have a `StructInfo` appended. Please come up with more guarantees or tests, that could make this more secure.


crates/mun_codegen/src/code_gen/symbols.rs
```diff
@@ -1,44 +1,64 @@
...
-            intern_string(module, &ty.name).into(),
-        ],
-        false,
-    )
+fn type_info_ir<D: IrDatabase>(
+    db: &D,
+    target: &TargetMachine,
+    module: &Module,
+    types: &AbiTypes,
+    type_values: &mut HashMap<TypeInfo, PointerValue>,
```
baszalmstra: Maybe rename to `type_info_to_ptr` or something to make it a little clearer

crates/mun_codegen/src/code_gen/symbols.rs
```diff
@@ -125,25 +148,60 @@ fn gen_signature_from_dispatch_entry<D: IrDatabase>(
...
-fn gen_type_info_array<D: IrDatabase>(
-    _db: &D,
+/// Recursively expands an array of TypeInfo into an array of IR pointers
+fn expand_type_info<D: IrDatabase>(
+    db: &D,
+    target: &TargetMachine,
     module: &Module,
     types: &AbiTypes,
+    type_values: &mut HashMap<TypeInfo, PointerValue>,
     hir_types: impl Iterator<Item = TypeInfo>,
```
baszalmstra: Rename to `type_info` or something like that

crates/mun_codegen/src/code_gen/symbols.rs
```diff
@@ -125,25 +148,60 @@ fn gen_signature_from_dispatch_entry<D: IrDatabase>(
...
+        gen_global(module, &type_array_ir, "").as_pointer_value()
+    }
+}
+
+/// Generates IR for an array of TypeInfo values
+fn gen_type_info_ptr_array(
+    module: &Module,
+    types: &AbiTypes,
+    type_values: &HashMap<TypeInfo, PointerValue>,
+    hir_types: impl Iterator<Item = TypeInfo>,
```
baszalmstra: Rename to `type_infos` or something like that.

crates/mun_codegen/src/code_gen/symbols.rs
Wodann: How about `global_type_info_map`?

PR #85, opened 2020-01-22
fix: change LineIndex::line_str() from using len() to chars().count()

legendiguess:


Wodann:
@legendiguess  Do you think you can add a unit test that detects the absence of the error this solves?

You can add it with the other unit test at the bottom of this file.

PR #87, opened 2020-01-24
Adds structs as a marshallable type

Wodann:
Depends on PR #84.

Allows getting, setting and replacing of structs in another struct:

```mun
let foo: Struct = some_foo_constructor();
let bar = foo.get<Struct>("bar");
let bar2 = some_bar_constructor();
let old_bar = foo.replace("bar", bar2);
foo.set("bar", old_bar);
```

### DIFF
--- a/crates/mun_abi/src/autogen_impl.rs
+++ b/crates/mun_abi/src/autogen_impl.rs
@@ -146,6 +146,22 @@ impl StructInfo {
             unsafe { slice::from_raw_parts(self.field_sizes, self.num_fields as usize) }
         }
     }
+
+    /// Returns the index of the field matching the specified `field_name`.
+    pub fn find_field_index(struct_info: &StructInfo, field_name: &str) -> Result<usize, String> {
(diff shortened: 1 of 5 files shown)


Wodann:
[Add marshalling of structs to Rust](https://trello.com/c/jEOrWCyF/116-add-marshalling-of-structs-to-rust)

crates/mun_abi/src/autogen_impl.rs
```diff
@@ -146,6 +146,22 @@ impl StructInfo {
             unsafe { slice::from_raw_parts(self.field_sizes, self.num_fields as usize) }
         }
     }
+
+    /// Returns the index of the field matching the specified `field_name`.
+    pub fn find_field_index(struct_info: &StructInfo, field_name: &str) -> Result<usize, String> {
```
baszalmstra: I think at some point we should figure out a way to do some sort of hash lookup. Because this might become slow.

PR #88, opened 2020-01-24
Add function to runtime CAPI for retrieving StructInfo from a TypeInfo pointer

Wodann:
Depends on #87.


Wodann:
[Add marshalling of structs to C++](https://trello.com/c/TYrIJt3i/117-add-marshalling-of-structs-to-c)

PR #89, opened 2020-01-24
Fix bug caused by erroneous caching of struct types

Wodann:
Depends on #88


Wodann:
[Solve struct hot reload bug](https://trello.com/c/3rSvdkmd/112-solve-struct-hot-reload-bug)

Wodann:
[Write tests for the C bindings](https://trello.com/c/wWRLqcXM/80-write-tests-for-the-c-bindings)

PR #91, opened 2020-02-05
feat(diagnostic): integrate annotate-snippets crate

legendiguess:


crates/mun_compiler/src/diagnostics.rs
```diff
@@ -3,111 +3,510 @@ use mun_hir::{FileId, HirDatabase, HirDisplay, Module};
...
 
-mod emit;
+use annotate_snippets::snippet::{Annotation, AnnotationType, Slice, Snippet, SourceAnnotation};
 
-pub use emit::Emit;
-use mun_errors::{Diagnostic, Level};
+use unicode_segmentation::UnicodeSegmentation;
+
+/// Replace all tabs('\t') in string to spaces(' ')
+fn replace_tabs(text: &mut String) {
```
baszalmstra: It would be more Rusty to insert a String and return a String instead of mutating the `text` param.

baszalmstra: Why not use: https://doc.rust-lang.org/std/primitive.str.html#method.replace ?

crates/mun_compiler/src/diagnostics.rs
```diff
@@ -3,111 +3,510 @@ use mun_hir::{FileId, HirDatabase, HirDisplay, Module};
...
+    let mut tab_index = text.find('\t');
+    while let Some(index) = tab_index {
+        #[allow(clippy::range_plus_one)]
+        text.replace_range(index..index + 1, " ");
+        tab_index = text.find('\t');
+    }
+}
+
+/// Returns amount of graphemes in text range
+fn graphemes_amount_in_range(range: (usize, usize), text: &str) -> usize {
```
baszalmstra: Maybe rename to `count_graphemes_in_range`

baszalmstra: Why not use `text[range.0, range.1]` ?

This entire function can be replaced with `UnicodeSegmentation::graphemes(text[range], true).count()` where range is the Range that you want to check.

See: https://doc.rust-lang.org/1.30.0/std/primitive.str.html#impl-Index%253CRange%253Cusize%253E%253E

crates/mun_compiler/src/diagnostics.rs
```diff
@@ -3,111 +3,510 @@ use mun_hir::{FileId, HirDatabase, HirDisplay, Module};
...
+    let (_, text) = text.split_at(range.0);
+    let (text, _) = text.split_at(range.1 - range.0);
+    UnicodeSegmentation::graphemes(text, true).count()
+}
+
+/// Constructs proper range for `Snippet` from `annotate-snippets` crate with taking
+/// folded text part in account, by subtraction first line offset from output range
+fn construct_range(
+    highliht_range: (usize, usize),
```
baszalmstra: Maybe use `Range<usize>` instead?

crates/mun_compiler/src/diagnostics.rs
```diff
@@ -3,111 +3,510 @@ use mun_hir::{FileId, HirDatabase, HirDisplay, Module};
...
+    text_part: &str,
+    first_line_offset: usize,
+) -> (usize, usize) {
```
baszalmstra: Maybe use `Range<usize>` instead?

crates/mun_compiler/src/diagnostics.rs
```diff
@@ -3,111 +3,510 @@ use mun_hir::{FileId, HirDatabase, HirDisplay, Module};
...
+        }
     }));
 
     let result = RefCell::new(result);
     let mut sink = DiagnosticSink::new(|d| {
-        result.borrow_mut().push(Diagnostic {
-            level: Level::Error,
-            loc: d.highlight_range().into(),
-            message: d.message(),
+        let first_line = line_index.line_col(d.highlight_range().start()).line;
```
baszalmstra: Definitly create more helper functions here, there is a LOT of code duplication in the creation of the snippets.

crates/mun_compiler/src/driver.rs
```diff
@@ -107,14 +106,22 @@ impl Driver {
...
             let diags = diagnostics(&self.db, file_id);
-            for diagnostic in diags.iter() {
-                diagnostic.emit(writer, &self.db, file_id)?;
-                if diagnostic.level == Level::Error {
-                    has_errors = true;
+            for diagnostic in diags {
+                let dl = DisplayList::from(diagnostic.clone());
+                eprintln!("{}", dlf.format(&dl));
+                if let Some(annotation) = diagnostic.title {
+                    #[allow(clippy::single_match)]
```
baszalmstra: I'd rather that you not ignore this clippy warning, but do as it says by using `if let`

crates/mun_compiler/src/driver.rs
```diff
@@ -107,14 +106,22 @@ impl Driver {
 
     /// Emits all diagnostic messages currently in the database; returns true if errors were
     /// emitted.
-    pub fn emit_diagnostics(&self, writer: &mut impl WriteColor) -> Result<bool, failure::Error> {
+    pub fn emit_diagnostics(&self) -> Result<bool, failure::Error> {
```
baszalmstra: Would it also be possible to still be able to pass a stream into this function to which the diagnostics are written.

legendiguess: I use `eprintln!` macro to output diagnostics, which write diagnostics in `io::stderr`. And so you can get this stream from anywhere by `io::stderr()`. Do that make sense?

legendiguess: > Would it also be possible to still be able to pass a stream into this function to which the diagnostics are written.

I use eprintln! macro to output diagnostics, which write diagnostics in io::stderr. And so you can get this stream from anywhere by io::stderr(). Do that make sense?

crates/mun_compiler/src/driver.rs
legendiguess: > I'd rather that you not ignore this clippy warning, but do as it says by using `if let`

I ignore this clippy warning because in the future here also can be matching with `AnnotationType::Warning`, `AnnotationType::Info` and etc. Do you anyway recommend for now change that matching to `if let`?

crates/mun_compiler/src/diagnostics.rs
legendiguess: > Maybe use `Range<usize>` instead?

`(usize, usize)` is what `annotate-snippets` api want for constructing `Snippet`.

baszalmstra:
[Create a crate to generate proper error messages and snippet from diagnostics](https://trello.com/c/ixhK7wq8/72-create-a-crate-to-generate-proper-error-messages-and-snippet-from-diagnostics)

crates/mun_compiler/src/diagnostics.rs
baszalmstra: Strange, but OK :)

crates/mun_compiler/src/driver.rs
baszalmstra: I would prefer that you pass in a stream and log to that instead of assuming that you want to output to stderr. So that the API becomes something like:
```
pub fn emit_diagnostics(&self, writer: &mut std::io::Write);
```
Something like that..

crates/mun/src/main.rs
```diff
@@ -136,6 +142,29 @@ fn compiler_options(matches: &ArgMatches) -> Result<mun_compiler::CompilerOption
         _ => return Err(format_err!("Only optimization levels 0-3 are supported")),
     };
 
+    let color = match matches.value_of("color") {
```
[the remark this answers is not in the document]
legendiguess: @baszalmstra  `value_of("color")` returns `&str`, but `env::var("MUN_TERM_COLOR")` returns `String`. I can't transform output from `env::var("MUN_TERM_COLOR")` to `&str` because of `returns a value referencing data owned by the current function`. Here's my current _not_ working code:
```
 let color = matches
        .value_of("color")
        .or_else(|| {
            env::var("MUN_TERMINAL_COLOR")
                .ok()
                .map(|value| value.as_str())
        })
        .map(|value| match value {
            "disable" => Color::Disable,
            "enable" => Color::Enable,
            _ => Color::Auto,
        })
        .unwrap_or(Color::Auto);
```
Any ideas?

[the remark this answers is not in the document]
baszalmstra: from the top of my head, first convert the `&str` to an owned string with `String::to_owned` :
```
matches
  .value_of("color")
  .map(ToOwned::to_owned)
  .or_else(|| env::var("MUN_TERM_COLOR")
  .map(|str| match &str { ... })
  .unwrap_or(Color::Auto)
```

crates/mun_compiler/src/driver/config.rs
```diff
@@ -14,6 +15,9 @@ pub struct Config {
     /// The optional output directory to store all outputs. If no directory is specified all output
     /// is stored in a temporary directory.
     pub out_dir: Option<PathBuf>,
+
+    /// Decide use or not to use colors in terminal output
```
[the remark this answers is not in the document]
legendiguess: >It still depends on the current terminal whether we actually do use colors.

We can't really be sure if current terminal support ansi colors, so color option is not "suggestion" is primary decree from user to compiler. So even if user terminal don't support ansi colors, `--color enable` turn them on.
Does that make sense?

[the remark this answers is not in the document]
baszalmstra: Yes, then it should be:
```suggestion
    /// Whether or not to use colors in terminal output
```

crates/mun_compiler/src/diagnostics_snippets.rs
```diff
@@ -0,0 +1,372 @@
...
+                .build(&source_code, source_code_len, &line_index),
+        )
+        .build();
+    // Add one to right range to make highlighting range here visible on output
+    snippet.slices[0].annotations[0].range.1 += 1;
+
+    snippet
+}
+
+pub(crate) fn error(
```
baszalmstra, line 73: Maybe rename to `generic_error`?

crates/mun_compiler/src/diagnostics_snippets.rs
```diff
@@ -0,0 +1,372 @@
...
+use mun_hir::{HirDatabase, HirDisplay};
+use mun_syntax::{
+    ast, AstNode, Parse, SourceFile, SyntaxError, SyntaxKind, SyntaxNodePtr, TextRange,
+};
+
+use std::sync::Arc;
+
+use mun_hir::line_index::LineIndex;
+
+use crate::annotate_snippets_builders::{AnnotationBuilder, SliceBuilder, SnippetBuilder};
```
Wodann, line 11: Maybe rename this to `crate::annotate` and include the `annotate_snippets::snippet::{AnnotationType, Snippet};` as `pub use`.

crates/mun_compiler/src/diagnostics_snippets.rs
```diff
@@ -0,0 +1,372 @@
...
+                .unwrap_or_else(|| syntax_node_ptr.range())
+        }
+        _ => syntax_node_ptr.range(),
+    }
+}
+
+pub(crate) fn syntax_error(
+    syntax_error: &SyntaxError,
+    _: &impl HirDatabase,
+    _: &Parse<SourceFile>,
```
Wodann, line 41: If not necessary, feel free to remove them.

crates/mun_compiler/src/diagnostics_snippets.rs
```diff
@@ -0,0 +1,372 @@
...
+    relative_file_path: &str,
+    source_code: &str,
+    source_code_len: usize,
+    line_index: &Arc<LineIndex>,
+) -> Snippet {
+    let mut snippet = SnippetBuilder::new()
+        .title(
+            AnnotationBuilder::new(AnnotationType::Error)
+                .label("syntax error".to_string())
```
Wodann, line 50: I'd prefer if the `label` function would accept `&str` instead of `String` only. That makes the API easier to use. Internally you can convert it to `String`. This one is also neat: https://doc.rust-lang.org/std/string/trait.ToString.html

crates/mun_compiler/src/diagnostics_snippets.rs
```diff
@@ -0,0 +1,372 @@
...
+                .build(),
+        )
+        .slice(
+            SliceBuilder::new(true)
+                .origin(relative_file_path.to_string())
```
Wodann, line 55: Ditto

crates/mun_compiler/src/diagnostics_snippets.rs
```diff
@@ -0,0 +1,372 @@
...
+                .source_annotation(
+                    (
+                        syntax_error.location().offset().to_usize(),
+                        syntax_error.location().end_offset().to_usize(),
+                    ),
+                    syntax_error.to_string(),
```
Wodann, line 61: Ditto.

crates/mun_compiler/src/diagnostics_snippets.rs
```diff
@@ -0,0 +1,372 @@
...
+    // Add one to right range to make highlighting range here visible on output
+    snippet.slices[0].annotations[0].range.1 += 1;
+
+    snippet
+}
+
+pub(crate) fn generic_error(
+    diagnostic: &dyn HirDiagnostic,
+    _: &impl HirDatabase,
+    _: &Parse<SourceFile>,
```
Wodann, line 76: Feel free to remove.

crates/mun_compiler/src/diagnostics_snippets.rs
```diff
@@ -0,0 +1,372 @@
...
+                    AnnotationType::Error,
+                )
+                .build(&source_code, source_code_len, &line_index),
+        )
+        .build()
+}
+
+pub(crate) fn unresolved_value_error(
+    diagnostic: &mun_hir::diagnostics::UnresolvedValue,
+    _: &impl HirDatabase,
```
Wodann, line 103: Ditto.

crates/mun_compiler/src/diagnostics_snippets.rs
```diff
@@ -0,0 +1,372 @@
...
+    parse: &Parse<SourceFile>,
+    relative_file_path: &str,
+    source_code: &str,
+    source_code_len: usize,
```
Wodann, line 107: A slice includes the length, so you should be able to incorporate this in `source_code: &str`

crates/mun_compiler/src/diagnostics_snippets.rs
```diff
@@ -0,0 +1,372 @@
...
+                    text_range_to_tuple(diagnostic.highlight_range()),
+                    format!("use of possibly-uninitialized `{}`", variable_name),
+                    AnnotationType::Error,
+                )
+                .build(&source_code, source_code_len, &line_index),
+        )
+        .build()
+}
+
+pub(crate) fn access_unknown_field_error(
```
Wodann, line 327: If there is this much code duplication, my intuition is that there should be a smarter/more concise way of doing this. I'd have to investigate to see what that method is.

crates/mun_compiler/src/driver.rs
```diff
@@ -1,27 +1,31 @@
...
+    snippet::{AnnotationType, Snippet},
+};
 
 pub const WORKSPACE: SourceRootId = SourceRootId(0);
 
 #[derive(Debug)]
 pub struct Driver {
     db: CompilerDatabase,
     out_dir: Option<PathBuf>,
+    color_option: Color,
```
Wodann, line 28: This makes me question the naming of the `Color` type. I'd expect it to be a colour, but it is actually an option

crates/mun_compiler/src/driver.rs
```diff
@@ -104,14 +109,25 @@ impl Driver {
...
-            for diagnostic in diags.iter() {
-                diagnostic.emit(writer, &self.db, file_id)?;
-                if diagnostic.level == Level::Error {
-                    has_errors = true;
+            for diagnostic in diags {
+                let dl = DisplayList::from(diagnostic.clone());
+                writeln!(writer, "{}", dlf.format(&dl)).unwrap();
+                if let Some(annotation) = diagnostic.title {
+                    #[allow(clippy::single_match)]
+                    match annotation.annotation_type {
```
Wodann, line 125: Either handle all cases explicitly, or make this an `if let AnnotationType::Error = annotation.annotation_type`. This is preferable imo because now you are implicitly hiding changes in the implementation. E.g. if someone were to add `AnnotationType::Fatal`, we would probably want to set `has_errors = true`. if someone were to read this now, they'd think: "probably the previous author skipped all other cases on purpose". By being explicit about all cases, the compiler will throw an error if in the future - if a new `AnnotationType` is added - or if you use `if let` someone reading the source code will notice that `AnnotationType::Fatal` is not explicitly handled.

crates/mun_compiler/src/driver/color.rs
```diff
@@ -0,0 +1,77 @@
+use std::env;
+#[cfg(target_os = "windows")]
+use std::process::Command;
+
+#[derive(Debug, Clone, Copy)]
+pub enum Color {
```
Wodann, line 6: Might need a more expressive name. Such as `DisplayColor`

crates/mun_compiler/src/driver/color.rs
```diff
@@ -0,0 +1,77 @@
...
+    pub(crate) fn should_enable(self) -> bool {
+        match self {
+            Color::Disable => false,
+            Color::Auto => terminal_support_ansi(),
+            Color::Enable => true,
+        }
+    }
+}
+
+/// Decides whether the current terminal supports ANSI escape codes basing on environment variable and user system.
```
Wodann, line 22:
```suggestion
/// Decides whether the current terminal supports ANSI escape codes based on the `term` environment variable and the operating system.
```

crates/mun_compiler/src/driver/color.rs
```diff
@@ -0,0 +1,77 @@
...
+            #[cfg(target_os = "windows")]
+            return windows_version_support_ansi();
+            #[cfg(not(target_os = "windows"))]
+            return false;
+        }
+    }
+}
+
+#[cfg(target_os = "windows")]
+/// Basing on the user version of Windows determines whether the Windows standart terminal support ANSI escape codes.
```
Wodann, line 39:
```suggestion
/// Determines whether the standard terminal supports ANSI escape codes, based on the user's version of Windows.
```

crates/mun_compiler/src/driver/color.rs
```diff
@@ -0,0 +1,77 @@
...
+
+                        Some((
+                            window_version[0].parse::<usize>(),
+                            window_version[1].parse::<usize>(),
+                            window_version[2].parse::<usize>(),
+                        ))
+                    });
+
+                if let Some((Ok(major), Ok(minor), Ok(patch))) = windows_version {
+                    // From Windows 10.0.10586 version and higher ANSI escape codes works in cmd
```
Wodann, line 66:
```suggestion
                    // From Windows 10.0.10586 version and higher ANSI escape codes work in `cmd`
```

crates/mun_hir/src/line_index.rs
```diff
@@ -38,17 +38,29 @@ impl LineIndex {
         }
     }
 
-    pub fn line_str<'a>(&self, line: u32, text: &'a str) -> Option<&'a str> {
-        let start_of_line = self.newlines.get(line as usize)?.to_usize();
-        let end_of_line = self
+    /// Get part of text between two lines
```
Wodann, line 41:
```suggestion
    /// Retrieves the text between `first_line` and `last_line`, if any.
```

crates/mun_hir/src/line_index.rs
```diff
@@ -38,17 +38,29 @@ impl LineIndex {
...
+            .get(last_line as usize + 1)
             .map(|u| u.to_usize() - 1)
-            .unwrap_or(text.len() as usize);
-        Some(&text[start_of_line..end_of_line])
+            .unwrap_or(text_len);
+        Some(&text[start_of_part..end_of_part])
     }
-}
 
+    /// Get offset to specific line
```
Wodann, line 58:
```suggestion
    /// Retrieves the offset to the line corresponding to `line_index`.
```

crates/mun_syntax/src/ast/extensions.rs
```diff
@@ -147,6 +147,25 @@ impl ast::StructDef {
     pub fn kind(&self) -> StructKind {
         StructKind::from_node(self)
     }
+    pub fn signature_range(&self) -> TextRange {
+        let sruct_kw = self
```
Wodann, line 151:
```suggestion
        let struct_kw = self
```

crates/mun_syntax/src/ast/extensions.rs
```diff
@@ -147,6 +147,25 @@ impl ast::StructDef {
...
+            .syntax()
+            .children_with_tokens()
+            .find(|p| p.kind() == T![struct])
+            .map(|kw| kw.text_range());
+        let name = self.name().map(|n| n.syntax.text_range());
+
+        let start = sruct_kw
```
Wodann, line 158:
```suggestion
        let start = struct_kw
```

crates/mun_syntax/src/ast/extensions.rs
```diff
@@ -147,6 +147,25 @@ impl ast::StructDef {
...
+            .map(|kw| kw.start())
+            .unwrap_or_else(|| self.syntax.text_range().start());
+
+        let end = name
+            .map(|name| name.end())
+            .or_else(|| sruct_kw.map(|kw| kw.end()))
```
Wodann, line 164:
```suggestion
            .or_else(|| struct_kw.map(|kw| kw.end()))
```

crates/mun_compiler/src/diagnostics.rs
```diff
@@ -1,113 +1,226 @@
...
     }));
 
     let result = RefCell::new(result);
     let mut sink = DiagnosticSink::new(|d| {
-        result.borrow_mut().push(Diagnostic {
-            level: Level::Error,
-            loc: d.highlight_range().into(),
-            message: d.message(),
-        });
+        result
```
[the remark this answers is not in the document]
baszalmstra: Maybe create an `ErrorContext` struct that contains all this information?

Wodann:
Just needs a squash before this can be merged. I will do that

Wodann:
Cleaned up history and rebase on `master`. Waiting for CI to succeed, then I will merge.

Wodann:
Removed changes to submodule hashes from respective commits, because this PR should really not be touching those. This was probably erroneously committed as a result of failing to `git submodule update`

PR #93, opened 2020-03-01
feat(runtime): add marshalling of value structs

Wodann:
@baszalmstra Can you check whether my approach is valid?

What is a good way to implement the generation of the wrapper? Should it be a copy of the body of the statically linked version, or should it just dereference the `struct(value)` parameters and pass it to the original function?


baszalmstra:
I would say we call the original function in the wrapper/bridge function.

Wodann:
[Write wrapper for value structs or stack allocated structs](https://trello.com/c/Ow7g4EQC/123-write-wrapper-for-value-structs-or-stack-allocated-structs)

PR #96, opened 2020-03-07
feat: extern functions

baszalmstra:
This PR adds the ability to define and call external functions that must be added to the runtime. E.g.

```mun
/// This is a function that does not have an implementation in Mun but instead is linked 
/// dynamically when loading the assembly
extern fn tick_the_system();

fn main() {
  // The extern functions can be called like any other Mun function
  tick_the_system()
}
```

When creating a runtime the `tick_the_system` function must be added to the runtime by calling `insert_fn`.

**Restrictions**

* Structs as parameters or return values are not yet supported


crates/mun_hir/src/expr/validator.rs
```diff
@@ -10,22 +15,70 @@ pub struct ExprValidator<'a, 'b: 'a, 'd, DB: HirDatabase> {
...
     pub fn validate_body(&mut self) {
         self.validate_uninitialized_access();
+        self.validate_extern();
+    }
+    pub fn validate_extern(&mut self) {
+        if !self.func.is_extern(self.db) {
+            return;
+        }
+
+        // Validate that there is not body
```
Wodann:
```suggestion
        // Validate that there is no body
```

crates/mun_runtime/src/function.rs
```diff
@@ -83,3 +84,51 @@ impl FunctionInfoStorage {
...
+
+pub trait IntoFunctionInfo {
+    fn into<S: AsRef<str>>(
+        self,
+        name: S,
+        privacy: abi::Privacy,
+    ) -> (FunctionInfo, FunctionInfoStorage);
+}
+
+impl<R: ReturnTypeReflection> IntoFunctionInfo for extern "C" fn() -> R {
```
Wodann: We should turn this into a macro similar to `invoke_fn_impl`, because now it only supports functions up to 2 arguments and with a lot of code duplication.

crates/mun_runtime/src/lib.rs
```diff
@@ -32,31 +32,56 @@ pub use crate::marshal::Marshal;
...
             options: RuntimeOptions {
                 library_path: library_path.into(),
                 delay: Duration::from_millis(10),
+                user_functions: Default::default(),
             },
-        }
+        };
+
+        result.insert_fn(
+            abi::Privacy::Private,
```
Wodann:
```suggestion
            abi::Privacy::Public,
```

crates/mun_runtime/src/lib.rs
```diff
@@ -32,31 +32,56 @@ pub use crate::marshal::Marshal;
...
+            "malloc",
+            malloc as extern "C" fn(u64, u64) -> *mut u8,
+        );
+
+        result.insert_fn(
+            abi::Privacy::Private,
```
Wodann:
```suggestion
            abi::Privacy::Public,
```

Wodann:
[Implement extern functions](https://trello.com/c/kjqRsD3U/54-implement-extern-functions)

crates/mun_runtime/src/function.rs
baszalmstra: Done

PR #97, opened 2020-03-13
Feature/object ptr indirection

Wodann:
Beware of several changes:
 1) target triple included in snapshots
 2) type info included in snapshots (as a result, target triples are also needed)
 3) inclusion of LLVM module in `#[salsa::input]` which might affect incremental compilation.

[IMPORTANT] To unblock PR #98, a resolution for 3) will be provided in a future PR!


crates/mun_abi/src/autogen_impl.rs
```diff
@@ -162,6 +162,13 @@ impl StructInfo {
                 )
             })
     }
+
+    /// Returns the size of the struct
+    pub fn size(&self) -> usize {
```
baszalmstra: Note that this might not be right, it could be that for cache efficiency there are padding bytes.

crates/mun_codegen/src/code_gen.rs
```diff
@@ -34,87 +39,122 @@ impl From<LinkerError> for CodeGenerationError {
...
+                &target.options.cpu,
+                &target.options.features,
+                db.optimization_lvl(),
+                RelocMode::PIC,
+                CodeModel::Default,
+            )
+            .ok_or(CodeGenerationError::CouldNotCreateTargetMachine)?;
+
+        // Initialize the module and target data
+        db.set_module(assembly_module.clone());
```
baszalmstra: This is basically the equivalent of setting global state which is very unsafe if the code is called from multiple threads at the same time. We have to find a better solution.

It might be better to NOT use salsa in this case because I don't think we're winning a lot with it in this case. Especially since the data we store in it is not truley clonable.

crates/mun_codegen/src/db.rs
```diff
@@ -14,6 +18,10 @@ pub trait IrDatabase: hir::HirDatabase {
     #[salsa::input]
     fn context(&self) -> Arc<Context>;
 
+    /// Returns the LLVM module that should be used for all generation steps.
+    #[salsa::input]
+    fn module(&self) -> Arc<Module>;
```
baszalmstra: This should definitely not be an input since its a transformation from HIR -> IR.

crates/mun_codegen/src/db.rs
```diff
@@ -22,6 +30,10 @@ pub trait IrDatabase: hir::HirDatabase {
     #[salsa::input]
     fn target(&self) -> Target;
 
+    /// Returns the target machine's data layout for code generation.
+    #[salsa::input]
+    fn target_data(&self) -> Arc<TargetData>;
```
baszalmstra: Same goes for this.

crates/mun_codegen/src/intrinsics.rs
```diff
@@ -17,8 +18,8 @@ pub trait Intrinsic: Sync {
 }
 
 intrinsics! {
-    /// Allocates memory from the runtime to use in code.
-    pub fn malloc(size: u64, alignment: u64) -> *mut u8;
+    /// Allocates memory for the specified type.
+    pub fn new(type: *const TypeInfo, alloc_handle: *mut ffi::c_void) -> *const *mut ffi::c_void;
```
baszalmstra: We should explain what this alloc_handle actually is. Even though I wrote it, its unclear to me when I read it.

crates/mun_codegen/src/ir/dispatch_table.rs
```diff
@@ -150,16 +154,39 @@ struct TypedDispatchableFunction {
...
             prototype_to_idx: Default::default(),
             entries: Default::default(),
             table_ref: None,
             table_type: module.get_context().opaque_struct_type("DispatchTable"),
+        };
+
+        if !intrinsics.is_empty() {
+            table.ensure_table_ref();
+
+            // Use a `BTreeMap` to guarantee deterministically ordered output
```
baszalmstra: Why is that important? A hashmap is also deterministic, right?

crates/mun_codegen/src/ir/type_table.rs
```diff
@@ -0,0 +1,325 @@
...
+    module::Module,
+    targets::TargetData,
+    values::{GlobalValue, IntValue, PointerValue, StructValue},
+    AddressSpace,
+};
+use std::collections::{BTreeSet, HashMap};
+use std::{mem, sync::Arc};
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct TypeTable {
```
baszalmstra: Missing comment, what is the TypeTable?

crates/mun_codegen/src/ir/type_table.rs
```diff
@@ -0,0 +1,325 @@
...
+    type_info_to_index: HashMap<TypeInfo, usize>,
+    entries: Vec<PointerValue>,
+    table_ref: PointerValue,
+}
+
+impl TypeTable {
+    pub fn gen_type_info_lookup(
```
baszalmstra: Missing docs: What does this call do exactly?

crates/mun_codegen/src/ir/body.rs
```diff
@@ -469,11 +496,21 @@ impl<'a, 'b, D: IrDatabase> BodyIrGenerator<'a, 'b, D> {
...
+                        .builder
+                        .build_load(value.into_pointer_value(), "mem_ptr")
+                        .into_pointer_value();
+
+                    self.builder.build_load(mem_ptr, "deref")
                 }
                 hir::StructMemoryKind::Value => {
                     if self.params.make_marshallable {
-                        self.builder.build_load(value.into_pointer_value(), "deref")
+                        let mem_ptr = self
```
baszalmstra: Duplicate code

baszalmstra:
So while looking a bit further, I found: http://llvm.org/doxygen/CloneModule_8cpp_source.html

This clearly shows what gets copied and what doesn't. I already found that you can re-lookup global variables and function definitions by name. Maybe we can write a little wrapper around modules, global values and function definitions that enables us to better clone these things?

crates/mun_abi/src/autogen_impl.rs
Wodann: If we take a struct:
```struct Foo {
  x: i16,
  y: u64
  _padding: [u8; 4],
}```

Both for the packed and unpacked version: the offset and size would be stored for `padding`, making the `size` function valid.

File: crates/mun_codegen/src/ir/dispatch_table.rs (hunk 6 above)
Comment (Wodann) [reply crates/mun_codegen/src/ir/dispatch_table.rs]: A hasmap doesn't guarantee order of elements, so every run results in different IR due to differently inserted dispatch table entries. This also affects the snapshot tests and in general you probably want a deterministic compile result.

File: crates/mun_codegen/src/ir/module.rs [hunk 10]
```diff
@@ -104,6 +119,19 @@ pub(crate) fn ir_query(db: &impl IrDatabase, file_id: FileId) -> Arc<ModuleIR> {
     let dispatch_table = dispatch_table_builder.finalize(&functions);
     let fn_pass_manager = function::create_pass_manager(&llvm_module, db.optimization_lvl());
 
+    // Create the allocator handle global value
+    let allocator_handle_global = if dispatch_table.has_intrinsic(&crate::intrinsics::new) {
```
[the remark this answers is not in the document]
Wodann: If `new` is there, `clone` must also be there - and vice versa.

crates/mun_abi/src/autogen_impl.rs
baszalmstra: I assume that LLVM also has something to say about the size of the struct regardless of whether you pad it yourself or not. But Im really not sure.

crates/mun_codegen/src/ir/module.rs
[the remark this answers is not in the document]
baszalmstra: Nonetheless if that is not the case in the future this will silently fail. Best to add it just in case.

crates/mun_abi/src/autogen_impl.rs
Wodann: Hmm. That would just be an offset between array elements, right? That doesn't affect the actual size of the type?
We could write a test for this, though - to be sure.

baszalmstra: Mmm maybe its only the case for alignment

```
struct A {
  a: u16,
  b: u8,
}
```

Would probably have an alignment of `u32`. WDYT?

crates/mun_codegen/src/ir/module.rs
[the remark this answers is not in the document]
Wodann: The `new`, `clone`, and - in the future - `drop` are all object management functions. They are all required if a struct type is defined. That should be the real test, I case because checking for everything potential future object management function doesn't scale too well.

crates/mun_codegen/src/ir/dispatch_table.rs
baszalmstra: :+1:

Wodann:
I don't think this will help the situation. The generation of module IR requires target data information. You mentioned that that is not allowed to be set in the salsa db, so we need to move module IR generation out of salsa. Thus there is no need for us to copy anything. We can see if we can move more _modular_ parts to salsa, though?

crates/mun_codegen/src/ir/module.rs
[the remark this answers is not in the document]
baszalmstra: I agree but this check is to determine whether or not a global variable allocator handle is required. This is only the case if either of these methods is required no matter where they are used. Maybe in the future we would like to use one of these functions for a different use case. Than we dont have to touch the check.

The best check would be to check if one of the functions in the dispatch table are actually using the AllocatorHandle.

baszalmstra:
> I don't think this will help the situation. The generation of module IR requires target data information. You mentioned that that is not allowed to be set in the salsa db, so we need to move module IR generation out of salsa. Thus there is no need for us to copy anything. We can see if we can move more _modular_ parts to salsa, though?

We can totally store the target in the database! Its not a function of other input right?

crates/mun_abi/src/autogen_impl.rs
Wodann: Yes, that's what I thought

crates/mun_codegen/src/db.rs
Wodann: I will resolve this issue in a future PR. I've changed the PR description to reflect this.

Wodann:
[Inject `Memory*` token into mun dll's and pass that to malloc calls](https://trello.com/c/etKWRreq/136-inject-memory-token-into-mun-dlls-and-pass-that-to-malloc-calls)

PR #101, opened 2020-03-22
fix(code_gen): incremental compilation

Wodann:
Resolves all three side-effects of PR #97.


Wodann:
> * IR of the group_ir currently doesnt have a regression test
What do you think is the best solution:
 * Output regression tests for the `group_ir` and `module_ir`?
 * Output a single regression test for the linked (and potentially optimized) IR

> Side note: did you try if incremental compilation works?
I don't currently have a solution for testing this. Do you have any ideas on how we could do it? This PR _merely_ restores things to the status quo before PR #97.

Wodann:
I've force pushed the requested additional comments. I also changed some of the file and struct names to better self-document, most notably `module::ModuleIR` -> `file::FileIR`. I felt this better reflects the changed structure.

I also [changed](https://github.com/mun-lang/mun/pull/101/files#diff-7a8bfdc626dbd0bcf8369b6b191559dfL245) the `file::FileIR`'s return type to no longer contain `FunctionValue`s, since those are only valid within the original LLVM `Module`, so it is unsafe to pass them around.

To faciliate this change, I also modified [this](https://github.com/mun-lang/mun/pull/101/files#diff-7a8bfdc626dbd0bcf8369b6b191559dfL166) and [that](https://github.com/mun-lang/mun/pull/101/files#diff-fbc6e7ac456f2304d189d1ff0dca5777R35).

baszalmstra:
> What do you think is the best solution:
> 
>     * Output regression tests for the `group_ir` **and** `module_ir`?
> 
>     * Output a single regression test for the linked (and potentially optimized) IR

Good question, I think both? But if I'd have to choose one, then probably the first.
 
> I don't currently have a solution for testing this. Do you have any ideas on how we could do it? This PR _merely_ restores things to the status quo before PR #97.

There is already one such test in https://github.com/mun-lang/mun/blob/master/crates/mun_hir/src/tests.rs . But I agree, maybe put that in a new PR. I do feel that it becomes more and more important as we go further. We want to make sure that we do not break incremental compilation. As we have seen that's very easy to do.

crates/mun_codegen/src/db.rs
```diff
@@ -42,9 +41,15 @@ pub trait IrDatabase: hir::HirDatabase {
...
+    /// TODO: Currently, a group always consists of a single file. Need to add support for multiple
+    /// files using something like `FileGroupId`.
+    #[salsa::invoke(crate::ir::file_group::ir_query)]
+    fn group_ir(&self, file: hir::FileId) -> Arc<FileGroupIR>;
+
     /// Given a `hir::FileId` generate code for the module.
-    #[salsa::invoke(crate::ir::module::ir_query)]
-    fn module_ir(&self, file: hir::FileId) -> Arc<ModuleIR>;
+    #[salsa::invoke(crate::ir::file::ir_query)]
+    fn module_ir(&self, file: hir::FileId) -> Arc<FileIR>;
```
baszalmstra: Maybe then rename this method to `file_ir` as well?

Wodann: Oops. Will do

PR #104, opened 2020-03-27
feat: performance benchmarks

baszalmstra:
This PR adds a few benchmarks to compare Mun to: Rust, LuaJIT and WASM. 

The PR also contains two performance optimizations.


crates/mun_runtime/benches/fibonacci.rs
```diff
@@ -0,0 +1,102 @@
...
+        group.bench_with_input(BenchmarkId::new("wasm", i), i, |b, i| {
+            b.iter(|| {
+                wasm.call("main", &[(*i as i32).into()]).unwrap();
+            })
+        });
+    }
+
+    group.finish();
+
+    fn fibonacci(n: i64) -> i64 {
```
Wodann: Why are there both a fibonacci.rs and these functions?

baszalmstra: Good catch, I removed the files in resources. I want to include the benchmark function in the benchmark itself to show what a reference implementation looks like.

Wodann: What is the source material for the *.wasm modules and simple instruction on how to generate them for new tests?

baszalmstra: I took the basic example: https://developer.mozilla.org/en-US/docs/WebAssembly/Rust_to_wasm. But instead of `greet` I implemented the functionality. But that could ofcourse be done with any language.

PR #108, opened 2020-03-30
fix: name of type table global

Wodann:


crates/mun_codegen/src/ir/type_table.rs
```diff
@@ -25,6 +25,8 @@ pub struct TypeTable {
 }
 
 impl TypeTable {
+    pub(crate) const NAME: &'static str = "global_type_table";
```
baszalmstra: This might be missing a comment. WDYT?

Wodann: I'll add one

Wodann:
Rebased on master, due to the recently merged PR. Waiting for checks to succeed.

PR #109, opened 2020-03-31
refactor: remove TypeInfo and StructInfo from StructRef

Wodann:
Instead we collect the most recent `TypeInfo` pointer from the `Runtime`'s
`GcRuntime`. This allows us to maintain the `StructRef` even after hot
reloads.

### DIFF
--- a/crates/mun_runtime/src/reflection.rs
+++ b/crates/mun_runtime/src/reflection.rs
@@ -1,14 +1,15 @@
 use crate::type_info::HasStaticTypeInfo;
-use crate::{marshal::Marshal, StructRef};
+use crate::{marshal::Marshal, Runtime, StructRef};
 use md5;
 
 /// Returns whether the specified argument type matches the `type_info`.
-pub fn equals_argument_type<'e, 'f, T: ArgumentReflection>(
+pub fn equals_argument_type<'r, 'e, 'f, T: ArgumentReflection>(
+    runtime: &'r Runtime,
     type_info: &'e abi::TypeInfo,
     arg: &'f T,
 ) -> Result<(), (&'e str, &'f str)> {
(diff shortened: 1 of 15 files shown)


crates/mun_runtime/src/struct.rs
```diff
@@ -155,7 +176,9 @@ impl ArgumentReflection for StructRef {
     type Marshalled = RawStruct;
 
     fn type_name(&self) -> &str {
-        self.info.name()
+        let runtime_ref = self.runtime.borrow();
+        let type_info = unsafe { &*runtime_ref.gc().ptr_type(self.handle.handle()).inner() };
+        type_info.name()
```
Wodann: This in particular worries me, because there is a significant performance overhead to retrieving a struct's type information. This change affects `StructRef::get`, `StructRef::replace`, and `invoke_fn!`.

It would already be a lot better if we didn't still have to `borrow` the `Runtime`. That means passing the `&Runtime` as an additional parameter to the `ArgumentReflection` trait's `type_name` and `type_guid` functions, though.

Wodann:
@baszalmstra I pushed a fix for the discussion about the runtime overhead of _borrowing_ a `Runtime` and rebased on `master`.

crates/mun_runtime/src/reflection.rs
```diff
@@ -90,32 +87,20 @@ macro_rules! impl_primitive_type {
...
+                    Self::type_info().name()
+                }
             }
         )+
     }
 }
 
 impl_primitive_type!(i8, i16, i32, i64, isize, u8, u16, u32, u64, usize, f32, f64, bool);
 
-impl ArgumentReflection for () {
```
Wodann, line 107: We currently don't support `()` (or empty) as an argument type, so I removed it.

PR #112, opened 2020-04-03
feature: early support for arrays

baszalmstra:
This PR adds simple early support for arrays. Arrays are always heap-allocated and have a dynamic size but cannot be resized at runtime (very similar to arrays in C#). 

**Syntax**

Array types are written as: `[T]`. Because in Mun all values have to be initialized we currently only support array construction by literal: `let a = [1,2,3,4,5,6,7,8,9,10]`.

Arrays will be indexable the same as in almost all languages with braces: `a[0]`. Indexing outside of the bounds of the array should result in a panic but that's outside of the scope of this PR (since we have no panic construct yet).

**GC support**

The GC stores array differently from other types. The data of the array is prepended with the size of the array. It will look something like this:

```
struct ArrayData {
   size_t numberOfElements;
   T elements[N];
}
```

**Progress**

- [x] Modify the ABI to support array types
- [x] Parsing of array types (`[int]`)
- [x] Support in the type system for arrays (`Ty::Array`)
- [x] Support for parsing array literals (`let a = [1,2,3]`)
- [x] HIR support for array literals
- [x] Parsing of array indexing
- [x] HIR support for array indexing
- [ ] Code generation for array types (passing them around generates correct IR)
- [ ] Code generation for array literals (allocation)
- [ ] Garbage collection support
- [ ] Code generation for array indexing


baszalmstra:
The development of this feature has been moved to `v0.4.0` (together with some other language features) because we want to focus on user/developer `experience` first.

baszalmstra:
Closing in favor of #341

PR #117, opened 2020-04-08
feat: memory mapping

Wodann:


crates/mun_hir/src/ty.rs
```diff
@@ -136,6 +136,18 @@ impl Ty {
             _ => None,
         }
     }
+
+    pub fn ty_to_string(&self, db: &impl HirDatabase) -> Option<String> {
```
baszalmstra, line 140: there already is something to do this `ty.display(db)`  will create a formatter.

crates/mun_runtime/src/assembly.rs
```diff
@@ -40,68 +42,131 @@ impl Assembly {
...
             library_path: library_path.to_path_buf(),
-            library: Some(library),
+            library,
+            legacy_libs: Vec::new(),
             info,
             allocator: gc,
-        })
+        };
+
+        assembly.ensure_linkable(runtime_dispatch_table)?;
```
baszalmstra, line 53: Wouldnt it be nicer to do this outside of the new call? That way you can also load an Assembly even though it doesn't link. We should obviously only take it in use if its actuall linkable but for the sake of single responsibility I think this call should be outside of the `new` call.

crates/mun_runtime/tests/util.rs
```diff
@@ -0,0 +1,106 @@
...
+            runtime: RuntimeOrBuilder::Builder(builder),
+        }
+    }
+
+    /// Updates the text of the Mun source and ensures that the generated assembly has been reloaded.
+    pub fn update(&mut self, text: &str) {
+        self.runtime_mut(); // Ensures that the runtime is spawned prior to the update
+        self.driver.set_file_text(self.file_id, text);
+        let out_path = self.driver.write_assembly(self.file_id).unwrap();
+        assert_eq!(
```
baszalmstra, line 67: We should output the compiler diagnostics here.

Wodann:
[Memory mapping](https://trello.com/c/P08jQOVP/142-memory-mapping)

PR #123, opened 2020-04-09
Use '->' instead of ':' for fn return types

### LINKED ISSUE #105: Replace return type annotation `:` with `->`
We would benefit from staying close to Rust syntax, however, currently return type annotations are written like:

```
fn foo(): int {
}
```

we would like to replace this syntax with this:

```
fn foo() -> int {
}
```

to be closer to Rust.

**Good first issue process:**

- [ ] implement the feature and start a Pull Request with in the description: closes #105. 

jakbyte:
Closes #105. I updated all of the tests but not the README example.


crates/mun_syntax/src/grammar.ron
```diff
@@ -48,6 +48,7 @@ Grammar(
         ["%=", "PERCENTEQ"],
         ["..=", "DOTDOTEQ"],
         ["::", "COLONCOLON"],
+        ["->", "MINUSGT"],
```
baszalmstra, line 51: I think `THIN_ARROW` might be a better name.

crates/mun_syntax/src/tests/snapshots/parser__function.snap
```diff
@@ -97,15 +98,23 @@ SOURCE_FILE@[0; 158)
...
+  ERROR@[148; 149)
+    COLON@[148; 149) ":"
+  ERROR@[149; 155)
+    IDENT@[149; 155) "number"
+  WHITESPACE@[155; 156) " "
+  ERROR@[156; 157)
+    L_CURLY@[156; 157) "{"
+  ERROR@[157; 158)
+    R_CURLY@[157; 158) "}"
+error Offset(120): expected a block
```
baszalmstra, line 110: These errors weren't here before. I think you forgot to change this test.

baszalmstra:
Also, why not change the README? that should reflect what the language looks like currently.

crates/mun_syntax/src/tests/snapshots/parser__function.snap
jakbyte: When I run `cargo test --release --all`, all of the tests succeed, and `cargo insta review` says that there are no snapshots to review.

crates/mun_syntax/src/grammar.ron
jakbyte: Okay, done!

crates/mun_syntax/src/tests/snapshots/parser__function.snap
baszalmstra: Thats correct because I think you forgot to change `:` to `->` in the input so the errors are to be expected. However, they weren't there before. I think you missed this file:
https://github.com/mun-lang/mun/blob/master/crates/mun_syntax/src/tests/parser.rs

jakbyte:
I'll update the `mun_runtime` example as well.

jakbyte:
It looks like I forgot to add the arrow to the `symbols` test, too.

crates/mun_syntax/src/tests/snapshots/parser__function.snap
jakbyte: Fixed.

jakbyte:
Did I miss anything else?

baszalmstra:
> 
> 
> Did I miss anything else?

I think you missed these : https://github.com/mun-lang/mun/tree/master/crates/mun_runtime/benches/resources

PR #129, opened 2020-04-12
perf: use a slab allocator for `ObjectInfo`

### LINKED ISSUE #106: Implement slab allocator to store `ObjectInfo<T>`
The `MarkSweep` struct implements `GcRuntime` to facilitate garbage collection for Mun. Currently, the implementation is quite inefficient. One of the things that could be improved is to implement a [slab allocator](https://en.wikipedia.org/wiki/Slab_allocation) (or object pool) instead of randomly allocating memory through a `Pin<Box<ObjectType<T>>>`. The general idea of the allocator is to allocate large chunks of memory to store instances of `ObjectType<T>` in and reusing them when they are deallocated. We can use a slab allocator here because, unlike the memory the `ObjectType<T>` is pointing to, the size of an `ObjectType<T>` is always the same.

Currently, Mun is single-threaded, however, this will not be the case forever. Therefore it is important to take thread-safety into account when designing the allocator. But probably anything is better than the current implementation.

**Good first issue process:**

- [ ] implement the feature and start a Pull Request with in the description: closes #106. 

foeb:
This should improve performance of the GC by limiting allocations of
`ObjectInfo` to large chunks.

Originally, I was hoping to use the allocator from the `slab` crate out
of the box, but that would cause all of the values to move when ever it
needed to resize. I added the `pinned_slab` module to get around that
limitation.

closes #106

### DIFF
--- a/crates/mun_memory/src/gc/mark_sweep.rs
+++ b/crates/mun_memory/src/gc/mark_sweep.rs
@@ -6,11 +6,10 @@ use crate::{
 };
 use once_cell::unsync::OnceCell;
 use parking_lot::RwLock;
+use pinned_slab::Slab;
 use std::{
-    collections::{HashMap, HashSet, VecDeque},
+    collections::{HashSet, VecDeque},
     hash::Hash,
-    ops::Deref,
-    pin::Pin,
 };
 
 /// Implements a simple mark-sweep type garbage collector.
@@ -20,7 +19,7 @@ where
     T: TypeLayout + TypeTrace + Clone,
     O: Observer<Event = Event>,
 {
-    objects: RwLock<HashMap<GcPtr, Pin<Box<ObjectInfo<T>>>>>,
+    objects: RwLock<Slab<ObjectInfo<T>>>,
     observer: O,
     stats: RwLock<Stats>,
 }
@@ -32,7 +31,7 @@ where
 {
     fn default() -> Self {
         MarkSweep {
-            objects: RwLock::new(HashMap::new()),
+            objects: RwLock::new(Slab::new()),
             observer: O::default(),
             stats: RwLock::new(Stats::default()),
         }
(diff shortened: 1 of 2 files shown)


crates/mun_gc/src/mark_sweep.rs
```diff
@@ -179,6 +173,61 @@ enum Color {
...
+
+impl<T: Type + Clone> MarkSweepObjects<T> {
+    pub fn new() -> Self {
+        MarkSweepObjects {
+            pool: Pool::new(),
+            info: HashMap::new(),
+        }
+    }
+
+    pub fn sweep(&mut self, observer: &impl Observer<Event = Event>, stats: &mut Stats) -> bool {
```
baszalmstra, line 203: I feel the code for this should reside somewhere in the `MarkSweep` object, where the tracing is also happening.

foeb:
@baszalmstra This is ready for review again :smile:

I moved the `Pool` code to another repo: https://github.com/foeb/pinned_slab

crates/mun_memory/src/gc/mark_sweep.rs
```diff
@@ -67,19 +66,20 @@ where
...
         // We want to return a pointer to the `ObjectInfo`, to be used as handle.
-        let handle = (object.as_ref().deref() as *const _ as RawGcPtr).into();
+        let handle: GcPtr;
 
         {
             let mut objects = self.objects.write();
-            objects.insert(handle, object);
+            let (_, ptr) = objects.insert(object);
+            handle = (ptr as *const ObjectInfo<T>).into();
         }
```
baszalmstra, line 83: You can return the handle directly from the block instead of setting it.
```let handle = { ... };```

File: crates/mun_memory/src/gc/mark_sweep.rs [hunk 3]
```diff
@@ -171,23 +168,26 @@ where
...
+                        .event(Event::Deallocation((obj as *const ObjectInfo<_>).into()));
                     stats.allocated_memory -= obj.ty.layout().size();
+                    false
                 }
-                false
-            }
-        });
+            })
+        };
+        objects.free_unused();
```
baszalmstra, line 190: What does this call do? Why is this not already done in the retain call?

Wodann:
Hi @foeb. I hope all is well.

@baszalmstra left some requested changes. Do you need additional information on those, or have you not had time to look at them yet?

If there is anything we can do to help out, please let us know

baszalmstra:
Im closing this since its become very stale.

PR #130, opened 2020-04-12
fix: fixes #128

### LINKED ISSUE #128: Calling an extern function without a return type exits the function it is called in
Calling an `extern fn` without a return type seems to exit the function it is called in.

For example, the following code

```rust
// resources/script.mun
extern fn thing(n: int);
extern fn print(n: int) -> int;

pub fn main() {
    // 1st
    print(1);
    thing(5);

    // 2nd
    print(2);
    thing(78);
}

// src/main.rs
use mun_runtime::{invoke_fn, RuntimeBuilder};
use std::{cell::RefCell, rc::Rc};

extern "C" fn thing(n: i64) {
    // do some cool stuff
}

extern "C" fn print(n: i64) -> i64 {
    println!("{}", n);
    // return an int so that it doesn't fail
    0
}

fn main() {
    let runtime = Rc::new(RefCell::new({
        let mut runtime = RuntimeBuilder::new("resources/script.munlib");
        runtime.insert_fn("thing", thing as extern "C" fn(i64));
        runtime.insert_fn("print", print as extern "C" fn(i64) -> i64);
        runtime.spawn().unwrap()
    }));

    let _: Result<_, mun_runtime::InvokeErr0<'_, ()>> = invoke_fn!(runtime, "main");
}
```

will output only

```
1
```

and not 2.

baszalmstra:
There was a bug where functions that returned nothing where seen as
functions that returned `never` during code generation. This caused dead
code elimination to remove the rest of the code.

I solved this issue by making sure that all functions do return
something. Unless the actually return type is `never`.

### DIFF
--- a/crates/mun_codegen/src/ir/body.rs
+++ b/crates/mun_codegen/src/ir/body.rs
@@ -218,7 +218,20 @@ impl<'a, 'b, D: IrDatabase> BodyIrGenerator<'a, 'b, D> {
                             .map(|expr| self.gen_expr(*expr).expect("expected a value"))
                             .collect();
 
-                        self.gen_call(def, &args).try_as_basic_value().left()
+                        self.gen_call(def, &args)
+                            .try_as_basic_value()
+                            .left()
+                            // If the called function is a void function it doesn't return anything.
+                            // If this method (`gen_expr`) returns None we assume the return value
+                            // is `never`. We return a const unit struct here to ensure that at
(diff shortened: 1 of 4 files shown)


baszalmstra:
> Would it be beneficial to add a `mun_runtime` test to verify that this actually also works for the generated code?

I've added a test to code generation where you can clearly see the IR being added for the calls. This was previously not the case. I think that's even better because you can see what happens.

Issue #133, opened 2020-04-13
Codegen Option::unwrap() panic during build


jakbyte:
The following script:

```rust
fn do_the_things(n: int) -> int {
    n + 7
}

pub fn main() {
    do_the_things(3);
}
```

causes this panic:

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', crates/mun_codegen/src/code_gen/symbols.rs:146:9
```

It compiles properly if `do_the_things` is `pub`.

It looks like it can't find the return type in the module for some reason.

baszalmstra:
Thanks for reporting, I will take a look when I have the time!

Wodann:
I am in the process of working on the fix. It involves a small refactor to include symbol generation in all codegen tests, which I will wrap up tomorrow.

PR #134, opened 2020-04-14
fix(codegen): only generate exposed functions

Wodann:
This also refactors codegen tests to always include symbol generation in
snapshot tests.

fixes issue #133

### DIFF
--- a/crates/mun_codegen/src/ir/file.rs
+++ b/crates/mun_codegen/src/ir/file.rs
@@ -25,13 +25,15 @@ pub(crate) fn ir_query(db: &impl IrDatabase, file_id: FileId) -> Arc<FileIR> {
 
     let group_ir = db.group_ir(file_id);
 
-    // Generate all function and wrapper function signatures.
+    // Generate all exposed function and wrapper function signatures.
     // Use a `BTreeMap` to guarantee deterministically ordered output.ures
     let mut functions = HashMap::new();
     let mut wrapper_functions = BTreeMap::new();
     for def in db.module_data(file_id).definitions() {
         if let ModuleDef::Function(f) = def {
-            if !f.is_extern(db) {
+            if (!f.data(db).visibility().is_private() || group_ir.dispatch_table.contains(*f))
+                && !f.is_extern(db)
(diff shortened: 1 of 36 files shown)


crates/mun_codegen/src/ir/type_table.rs
```diff
@@ -136,7 +142,10 @@ impl<'a, D: IrDatabase> TypeTableBuilder<'a, D> {
     /// Collects unique `TypeInfo` from the specified function signature and body.
     pub fn collect_fn(&mut self, hir_fn: hir::Function) {
         // Collect type info for exposed function
-        if !hir_fn.data(self.db).visibility().is_private() || hir_fn.is_extern(self.db) {
+        if !hir_fn.data(self.db).visibility().is_private()
```
baszalmstra, line 145: Are some of these options obsolete now? As in, if a function is externally visible is it not by definition in the dispatch table?

Wodann: Yup. You are right. I'll fix that.

crates/mun_codegen/src/ir/file.rs
```diff
@@ -25,13 +25,15 @@ pub(crate) fn ir_query(db: &impl IrDatabase, file_id: FileId) -> Arc<FileIR> {
...
 
-    // Generate all function and wrapper function signatures.
+    // Generate all exposed function and wrapper function signatures.
     // Use a `BTreeMap` to guarantee deterministically ordered output.ures
     let mut functions = HashMap::new();
     let mut wrapper_functions = BTreeMap::new();
     for def in db.module_data(file_id).definitions() {
         if let ModuleDef::Function(f) = def {
-            if !f.is_extern(db) {
+            if (!f.data(db).visibility().is_private() || group_ir.dispatch_table.contains(*f))
```
baszalmstra, line 34: Is it the same here?

Wodann: No, here we on purpose don't want to avoid the extern functions and private functions that are dead code.

PR #135, opened 2020-04-14
feat: remainder operator

jakbyte:
Adds the remainder operators `%` and `%=`. Please tell me if I missed anything. :slightly_smiling_face:


Wodann:
Nice work! The PR looks good overall.

I had one question, does the _signed remainder_ do the same thing as Rust, e.g. `-2 % 5 == -2`? I tried looking up the apropiate LLVM docs, but couldn't easily find them (on mobile)..

jakbyte:
I got a panic when I tried to test that :grimacing:

jakbyte:
Never mind, it works when I'm not trying to pass it into an `extern fn`.
I'll try to fix it.

jakbyte:
Wait, no. It doesn't work when I use it in a `pub fn` :grimacing:

jakbyte:
Here's the script:

```rust
pub fn main() {
    let x = -2 % 5;
}
```

[Here's the backtrace.](https://gist.github.com/jakbyte/c0a7e649a9efb2f2a872276d8f98cb90)

It looks like the error starts [here](https://github.com/jakbyte/mun-fork/blob/rem-op/crates/mun_codegen/src/ir/body.rs#L580) with the LHS' type being unknown. Maybe it has something to do with the unary minus operator?

jakbyte:
Yeah, it's probably the unary minus operator, as this same error happens with other operators, and it doesn't which side the negative number is on.

jakbyte:
Fixing that's outside the scope of this PR, though ¯\\\_(ツ)_/¯

###### (sorry for the wall of text :sweat_smile:)

jakbyte:
I implemented unary operator IR generation anyways, but the type is still unknown for whatever reason.

jakbyte:
Anyways, this should be working. That unary op stuff doesn't have anything to do with this.

crates/mun_codegen/src/ir/body.rs
```diff
@@ -734,6 +734,7 @@ impl<'a, 'b, D: IrDatabase> BodyIrGenerator<'a, 'b, D> {
             ArithOp::Subtract => self.builder.build_int_sub(lhs, rhs, "sub"),
             ArithOp::Divide => self.builder.build_int_signed_div(lhs, rhs, "div"),
             ArithOp::Multiply => self.builder.build_int_mul(lhs, rhs, "mul"),
+            ArithOp::Remainder => self.builder.build_int_signed_rem(lhs, rhs, "rem"),
```
baszalmstra, line 737: I think this should differ for signed and unsigned integers: build_int_unsigned_rem

baszalmstra: probably the same needs to happen for div, now that I look at it

baszalmstra:
Maybe you can add a test like Wodann mentioned using this:

```rust
pub fn main() {
    let x = (0-2) % 5;
}
```

Im missing a few tests to see if the output actually makes sense (I dont exactly know what LLVM does)

PR #136, opened 2020-04-15
feat: unary operators

jakbyte:
I think I've gotten codegen done, but it seems that the type of expressions with unary operators is, for whatever reason, unknown.

### DIFF
--- a/crates/mun_codegen/src/ir/body.rs
+++ b/crates/mun_codegen/src/ir/body.rs
@@ -5,7 +5,7 @@ use crate::{
 };
 use hir::{
     ArenaId, ArithOp, BinaryOp, Body, CmpOp, Expr, ExprId, HirDisplay, InferenceResult, Literal,
-    Name, Ordering, Pat, PatId, Path, Resolution, Resolver, Statement, TypeCtor,
+    Name, Ordering, Pat, PatId, Path, Resolution, Resolver, Statement, TypeCtor, UnaryOp,
 };
 use inkwell::{
     builder::Builder,
@@ -206,6 +206,7 @@ impl<'a, 'b, D: IrDatabase> BodyIrGenerator<'a, 'b, D> {
             Expr::BinaryOp { lhs, rhs, op } => {
                 self.gen_binary_op(expr, *lhs, *rhs, op.expect("missing op"))
             }
+            Expr::UnaryOp { expr, op } => self.gen_unary_op(*expr, *op),
             Expr::Call {
                 ref callee,
                 ref args,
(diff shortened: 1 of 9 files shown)


baszalmstra:
Seems to be because this is left unimplemented: 

https://github.com/mun-lang/mun/blob/39babc45c86cdb7232f077cf8b79f7caeda3e24b/crates/mun_hir/src/ty/infer.rs#L403

crates/mun_codegen/src/ir/body.rs
```diff
@@ -585,6 +586,61 @@ impl<'a, 'b, D: IrDatabase> BodyIrGenerator<'a, 'b, D> {
         }
     }
 
+    /// Generates IR to calculate a unary operation on an expression.
+    fn gen_unary_op(&mut self, expr: ExprId, op: UnaryOp) -> Option<BasicValueEnum> {
+        let ty = self.infer[expr].clone();
+        match ty.as_simple() {
+            Some(TypeCtor::Float(_ty)) => self.gen_unary_op_float(expr, op),
+            Some(TypeCtor::Int(ty)) => self.gen_unary_op_int(expr, op, ty.signedness),
```
baszalmstra, line 594: I think its important to ensure that the `Int` type is unsigned. You can match on the signedness of `ty`.

crates/mun_codegen/src/ir/body.rs
```diff
@@ -585,6 +586,61 @@ impl<'a, 'b, D: IrDatabase> BodyIrGenerator<'a, 'b, D> {
...
+        expr: ExprId,
+        op: UnaryOp,
+        _signedness: hir::Signedness,
+    ) -> Option<BasicValueEnum> {
+        let value: IntValue = self
+            .gen_expr(expr)
+            .map(|value| self.opt_deref_value(self.infer[expr].clone(), value))
+            .expect("no value")
+            .into_int_value();
+        match op {
```
baszalmstra, line 625: Checking the signedness could also happen here

jakbyte:
Looks like CI is failing due to a network error.

crates/mun_hir/src/ty/infer.rs
```diff
@@ -400,8 +400,7 @@ impl<'a, D: HirDatabase> InferenceResultBuilder<'a, D> {
                     }
                 }
             }
-            Expr::UnaryOp { .. } => Ty::Unknown,
-            //            Expr::UnaryOp { expr: _, op: _ } => {}
+            Expr::UnaryOp { expr, .. } => self.infer_expr(*expr, &Expectation::none()),
```
baszalmstra, line 403: This would also allow:

```
struct Foo;
fn main() {
  let a = Foo;
  let b = -Foo;
}
```

which doesn't make sense

baszalmstra: You probably want to check the type of the `infer_expr_inner` call and based on that determine the return type as well as emitting an error for types that cannot be negated or notted. See the [rust analyzer source](https://github.com/rust-analyzer/rust-analyzer/blob/f1a07dbf5559e882f46e79ed2a299cf151b99498/crates/ra_hir_ty/src/infer/expr.rs#L313) for inspiration (we used that a lot ;) )

crates/mun_hir/src/ty/snapshots/tests__invalid_unary_ops.snap
```diff
@@ -0,0 +1,15 @@
+---
+source: crates/mun_hir/src/ty/tests.rs
+expression: "fn bar(a: int, b: bool) {\n    a = !a; // not is not implemented for int\n    b = -b; // neg is not implemented for bool\n}"
```
baszalmstra, line 3: This does not seem to match with the expected output. There are no errors reported.

crates/mun_codegen/src/ir/body.rs
jakbyte: Should it return `None`, or should it just use `unimplemented!`?

baszalmstra: Panic might be better, its invalid to ever get there.

crates/mun_hir/src/ty/infer.rs
jakbyte: Should the signedness of ints be checked here as well?

jakbyte:
It looks like not is actually implemented for integers in Rust, so I'll go ahead and add that.

Wodann:
I checked master, but the `mun_abi/c` submodule is currently set to `b437e21`. You updated it to the commit of @baszalmstra's WIP array support, but since that is not in master yet, the abi should not be updated to that commit yet.

crates/mun_codegen/src/ir/body.rs
```diff
@@ -585,6 +586,67 @@ impl<'a, 'b, D: IrDatabase> BodyIrGenerator<'a, 'b, D> {
...
+        op: UnaryOp,
+        signedness: hir::Signedness,
+    ) -> Option<BasicValueEnum> {
+        let value: IntValue = self
+            .gen_expr(expr)
+            .map(|value| self.opt_deref_value(self.infer[expr].clone(), value))
+            .expect("no value")
+            .into_int_value();
+        match op {
+            UnaryOp::Neg => {
```
baszalmstra, line 626: The `not` operator is also implemented for integers in HIR but not here.

crates/mun_hir/src/ty/infer.rs
```diff
@@ -400,9 +400,23 @@ impl<'a, D: HirDatabase> InferenceResultBuilder<'a, D> {
...
-            Expr::UnaryOp { .. } => Ty::Unknown,
-            //            Expr::UnaryOp { expr: _, op: _ } => {}
-            //            Expr::Block { statements: _, tail: _ } => {}
+            Expr::UnaryOp { expr, op } => {
+                let ty = self.infer_expr(*expr, &Expectation::none());
+                if let Some(simple) = ty.as_simple() {
+                    match op {
+                        UnaryOp::Not => match simple {
+                            TypeCtor::Bool | TypeCtor::Int(_) => ty,
+                            _ => Ty::Unknown,
```
baszalmstra, line 409: There should be an error emitted here. Something similar to binary operation error: ```cannot apply operator `-` on type `int`.```

crates/mun_hir/src/ty/infer.rs
```diff
@@ -400,9 +400,23 @@ impl<'a, D: HirDatabase> InferenceResultBuilder<'a, D> {
...
+                        },
+                        UnaryOp::Neg => match simple {
+                            TypeCtor::Float(_) | TypeCtor::Int(_) => ty,
+                            _ => Ty::Unknown,
```
baszalmstra, line 413: ditto

jakbyte: ?

baszalmstra: Same as the above comment:

There should be an error emitted here. Something similar to binary operation error: cannot apply operator `-` on type `int`.

crates/mun_hir/src/ty/infer.rs
jakbyte: So a panic?

baszalmstra: See line 308 of this same file. There is a `Diagnostic` added: `CannotApplyBinaryOp`. There should be something similar with a unary operator.

crates/mun_hir/src/ty/infer.rs
jakbyte: Okay.

crates/mun_codegen/src/ir/body.rs
```diff
@@ -585,6 +586,68 @@ impl<'a, 'b, D: IrDatabase> BodyIrGenerator<'a, 'b, D> {
...
+                if signedness == hir::Signedness::Signed {
+                    Some(self.builder.build_int_neg(value, "neg").into())
+                } else {
+                    unimplemented!("Operator {:?} is not implemented for unsigned integer", op)
+                }
+            }
+            UnaryOp::Not => Some(self.builder.build_not(value, "not").into()),
```
baszalmstra, line 633: There is a test missing for this functionality

crates/mun_codegen/src/test.rs
```diff
@@ -92,6 +92,21 @@ fn binary_expressions() {
     );
 }
 
+#[test]
+fn unary_expressions() {
+    test_snapshot(
+        r#"
+    pub fn neg(x: int) -> int {
```
baszalmstra, line 99: I think its important to also test:
* not int
* neg float

jakbyte:
Sorry about this, but I (still) don't know how to rebase. Can you give me some steps on doing that?

baszalmstra:
Sure, does this help: https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request

Squashing your changes is not necessary.

PR #140, opened 2020-04-23
feat: add casting of fundamental types to memory mapping

Wodann:
Adds support for fundamental type conversions, when memory mapping. E.g. going from this struct:

```mun
struct Foo(
    u8,
    i16,
    u32,
    i64,
    f32,
)
```

to the following struct, is legal:
```mun
struct Foo(
    u16,
    i32,
    u64,
    i128,
    f64,
)
```
[Discussion] Non-guaranteed type conversions will result in ~~a `panic`~~ zero initialization. Is this desirable? Potentially we could at least attempt the conversion and only ~~`panic!`~~ zero initialize if the `try_into` call fails.

[Discussion] ~~Non-legal or failed type conversions will result in a `panic!`. Is this desirable?~~ Alternatives are:
1) Logging an error and initialising to zero.
2) Graciously fail the linking of the `Assembly`. Changes will have to be made to the source code before we re-evaluate the `Assembly`. This means that you can never "reuse" a field by doing an illegal conversion. This might force people to take a two-step hot reloading approach: i.e. remove first and then add the new field.

Edit: after discussion, we decided to never panic. If we cannot resolve the type conversion, we continue with the mapping - assuming zero initialization.


Wodann:
@baszalmstra any comments on the discussion topics in this description?

baszalmstra:
> [Discussion] Non-guaranteed type conversions will result in a panic. Is this desirable? Potentially we could at least attempt the conversion and only panic! if the try_into call fails.

I'd say that it should never `panic!` because I user can never catch that. 

[Discussion] Non-legal or failed type conversions will result in a panic!. Is this desirable? 

No, I don't think we should ever `panic!`. I think this is something that we have to play with before we can really make a decision. So far I think option number 1 is the most user friendly.

Wodann:
I'll add a commit to init to zero, in that case 👌🏻

Wodann:
I've updated the PR description to reflect that we decided to never `panic!` and added a separate commit to implement and test this behavior.

Could you review that commit please @baszalmstra ?

PR #142, opened 2020-04-28
cleanup runtime/memory code

Wodann:
Depends on #140.

### DIFF
--- a/crates/mun_memory/src/gc/mark_sweep.rs
+++ b/crates/mun_memory/src/gc/mark_sweep.rs
@@ -1,14 +1,13 @@
 use crate::{
     cast,
-    diff::{Diff, FieldDiff},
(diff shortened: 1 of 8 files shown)


baszalmstra:
Looks nice, one questions: Would it be possible to add unsafe `cast`ing when hotloading? From `u128` to `u64` for example?

Wodann:
> Looks nice, one questions: Would it be possible to add unsafe `cast`ing when hotloading? From `u128` to `u64` for example?

@baszalmstra Yes, as I mentioned in the first discussion topic of this PR #140, we can add *non-guaranteed* (or *unsafe*) casts, but when those fail they would result in zeroed memory. Would you be happy with that? If so, I'll add it to PR #140 or create a new PR for that specific case.

baszalmstra:
If a warning is emitted for that as well, that would be totally fine by me! :)

PR #144, opened 2020-04-29
feat: add assignment, bit, and boolean operators

Wodann:
This PR adds:

- [x] `bool` assignment operator
- [x] `struct` assignment operator (both for `struct(value)` and `struct(gc)`)
- [x] `bool` and/or (`a && b || c`)
- [x] bitwise and/or/xor operations for all integer types and `bool` (`a | b & c`, `a |= b`, `a &= b`)
- [x] left-shift and right-shift (-assignment) operations for all integer types (` a >> b`, `a << b`, `a >>= b`, `a <<= b`)

Tasks that were not done yet, but should be in a future PR:
- [ ] Add tests for operator errors
- [ ] left-shift and right-shift operations should work for all combinations of integer types. Currently it only works when the same integer type is used on the lhs and rhs of the expression. This is currently impossible as a result of [this](https://github.com/mun-lang/mun/commit/518f1d19d80a1ad1526054029cef7bcba7b3ff3a#diff-95270f11e9c60cb360f2477d2bdb0961R300) `Diagnostic`

**[discussion]** ~~Do we want to use the `^`  sign for bitwise xor or the power operator?~~ We've decided to use the `^`  sign for bitwise xor


Wodann:
I agree. In that case I will add `^` and `^=` for xor.

Wodann:
I've added the bitwise xor operator as well as the left-shift and right-shift operators. For scrutiny, could you please check the changed files @baszalmstra ?

baszalmstra:
Still looks good to me!

PR #154, opened 2020-05-01
feat: implements number type inferencing

### LINKED ISSUE #118: Add int and float literal type inference
Currently, the compiler sees all integer literals as `int` and all float literals as `float`, which means that the following code:

```rust
fn add(x: u8): u8 {
    x + 1
}

fn main() {
    let four = add(3);
}
```

will produce the following error:

```
error: expected `u8`, found `int`
 --> test.mun:2:8
  |
2 |     x + 1
  |         ^
error: expected `u8`, found `int`
 --> test.mun:2:4
  |
2 |     x + 1
  |     ^^^^^
error: expected `u8`, found `int`
 --> test.mun:1:18
  |
1 | fn add(x: u8): u8 {
  |
error: expected `u8`, found `int`
 --> test.mun:6:19
  |
6 |     let four = add(3);
  |       
```

(a similar error occurs when `f32` is used in place of `u8`)

baszalmstra:
Implements type inferencing of number types.

Closes #118

### DIFF
--- a/crates/mun_hir/src/ty/infer/type_variable.rs
+++ b/crates/mun_hir/src/ty/infer/type_variable.rs
@@ -1,9 +1,6 @@
-use crate::Ty;
-use drop_bomb::DropBomb;
-use ena::snapshot_vec::{SnapshotVec, SnapshotVecDelegate};
-use ena::unify::{InPlace, InPlaceUnificationTable, NoError, UnifyKey, UnifyValue};
-use std::borrow::Cow;
-use std::fmt;
+use crate::{ty::infer::InferTy, ty_app, Ty, TypeCtor};
+use ena::unify::{InPlaceUnificationTable, NoError, UnifyKey, UnifyValue};
+use std::{borrow::Cow, fmt};
 
 /// The ID of a type variable.
 #[derive(Copy, Clone, PartialEq, Eq, Hash, Debug)]
@@ -78,7 +75,6 @@ impl UnifyValue for TypeVarValue {
 
 #[derive(Default)]
 pub struct TypeVariableTable {
-    values: SnapshotVec<Delegate>,
     eq_relations: InPlaceUnificationTable<TypeVarId>,
 }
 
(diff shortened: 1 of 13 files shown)


crates/mun_hir/src/ty/infer/type_variable.rs
```diff
@@ -121,106 +190,112 @@ impl TypeVariableTable {
...
-        // `SnapshotVec` entirely, except that we would have to
-        // reproduce *some* of its logic, since we want to know which
-        // type variables have been instantiated since the snapshot
-        // was started, so we can implement `types_escaping_snapshot`.
-        //
-        // (If we extended the `UnificationTable` to let us see which
-        // values have been unified and so forth, that might also
-        // suffice.)
-    }
+    // /// Returns indices of all variables that are not yet instantiated.
```
Wodann, line 278: Is this for future reference?

crates/mun_hir/src/utils.rs
```diff
@@ -0,0 +1,10 @@
+use std::sync::Arc;
+
+/// Helper for mutating `Arc<[T]>` (i.e. `Arc::make_mut` for Arc slices).
+/// The underlying values are cloned if there are other strong references.
+pub(crate) fn make_mut_slice<T: Clone>(a: &mut Arc<[T]>) -> &mut [T] {
+    if Arc::get_mut(a).is_none() {
+        *a = a.iter().cloned().collect();
+    }
+    Arc::get_mut(a).unwrap()
```
Wodann, line 9:
```suggestion
    if let Some(a) = Arc::get_mut(a) {
        a
    } else {
        *a = a.iter().cloned().collect();
    }
```

crates/mun_hir/src/ty/infer/type_variable.rs
baszalmstra: Yes, I left it in there because it seems useful in the future.

crates/mun_hir/src/utils.rs
baszalmstra: We need to return the value as well. So `*a = a.iter().cloned().collect();` doesnt work. Since we know we are the ownly owners at that point we can safely unwrap.

Wodann: Never mind. I missed the assignment to the original `&mut Arc`.

PR #155, opened 2020-05-01
fix: implements binary comparison operator for bool

baszalmstra:
Implements missing binary comparison operator

### DIFF
--- a/crates/mun_codegen/src/test.rs
+++ b/crates/mun_codegen/src/test.rs
@@ -210,7 +210,21 @@ macro_rules! test_number_operator_types {
                         "#, ty = stringify!($ty),
                     ));
                 }
+            }
+        )+
+    };
+}
 
+test_number_operator_types!(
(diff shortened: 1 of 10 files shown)


Wodann:
Is there a reason not to include the other comparative operators?

baszalmstra:
> Is there a reason not to include the other comparative operators?
I dont feel like its intuitive, is `true` > `false`?

Wodann:
> I dont feel like its intuitive, is `true` > `false`?

Yes. At first I thought it might be a useless operator, but on second thought I realised those operators are necessary when ordering `bool` or when making an algorithm that allows transitions from `false -> false`, `false -> true`, and `true -> true` when running a loop, but no "reversions" (i.e. `true -> false`)

crates/mun_codegen/src/test.rs
```diff
@@ -125,6 +125,17 @@ fn assignment_op_bool() {
     );
 }
 
+#[test]
```
Wodann, line 128: This test is no longer necessary

PR #156, opened 2020-05-01
fix: comparison not implemented for structs

baszalmstra:
The comparison operator is not implemented for structs. This should be reflected in HIR.

### DIFF
--- a/crates/mun_hir/src/ty/snapshots/tests__comparison_not_implemented_for_struct.snap
+++ b/crates/mun_hir/src/ty/snapshots/tests__comparison_not_implemented_for_struct.snap
@@ -1,0 +1,9 @@
+---
+source: crates/mun_hir/src/ty/tests.rs
+expression: "struct Foo;\n\nfn main() -> bool {\n    Foo == Foo\n}"
+---
+[37; 47): cannot apply binary operator
+[31; 49) '{     ... Foo }': bool
+[37; 40) 'Foo': Foo
+[37; 47) 'Foo == Foo': bool
+[44; 47) 'Foo': Foo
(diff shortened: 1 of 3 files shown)


crates/mun_hir/src/ty/tests.rs
```diff
@@ -8,6 +8,18 @@ use mun_syntax::{ast, AstNode};
 use std::fmt::Write;
 use std::sync::Arc;
 
+#[test]
+fn comparison_not_implemented_for_struct() {
```
Wodann, line 12: Where is the corresponding `insta` file?

baszalmstra: Woops! Added

crates/mun_hir/src/ty/snapshots/tests__comparison_not_implemented_for_struct.snap
```diff
@@ -0,0 +1,9 @@
+---
+source: crates/mun_hir/src/ty/tests.rs
+expression: "struct Foo;\n\nfn main() -> bool {\n    Foo == Foo\n}"
+---
+[37; 47): cannot apply binary operator
+[31; 49) '{     ... Foo }': bool
+[37; 40) 'Foo': Foo
+[37; 47) 'Foo == Foo': bool
```
Wodann, line 8: It's interesting that it still knows what type inference results in, even though it is *illegal*

baszalmstra: That because applying the equality operator will always have a boolean as return type no matter the left- and right-hand types.

Wodann: Gotcha

PR #159, opened 2020-05-02
feature: compiler fuzzing

tdejager:
Adds fuzzing for AFL and rust fuzz to the project


baszalmstra:
Im hesitant to merge this because I dont see the added value of the current implementation. I've run the suite for multiple days and nothing was reported. Im doubting if we will ever find something with what it currently does..

WDYT?

Wodann:
@tdejager your thoughts?

@baszalmstra what would be next steps for this PR, in that case? A keyword-aware implementation of fuzzing?

baszalmstra:
Maybe there is something to be learned from this? 

https://blog.trailofbits.com/2020/06/05/breaking-the-solidity-compiler-with-a-fuzzer/

@wodann I think the current impl is already keyword aware. 

Im hesitant to close this too, because I dont want this work to go to waste. I also dont think it should live in master.

tdejager:
I agree we shouldn't merge this, and it is keyword aware, at least it is aware of the keywords. And use these to perturb the input.

Closing this would not necessarily let the work go to waste, we just now it doesn't work this way :)

baszalmstra:
Closing this for now

PR #161, opened 2020-05-02
misc: updated readme

baszalmstra:
Updates the readme with a working example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,16 +50,15 @@
 
 ```mun
 fn main() {
-    let sum = add(a, b);
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -50,16 +50,15 @@
 
 ```mun
 fn main() {
-    let sum = add(a, b);
-
-    // Comments: Mun natively supports bool, float, and int
+    // Comments: Mun natively supports bool, f32, f64, i8, u8, i16, etc, etc.
```
jakbyte, line 53: There are two `etc`'s here.

PR #171, opened 2020-05-08
improvement(memory): full support for struct memory mapping

Wodann:
In addition to previous available struct memory mapping, this PR adds:

* field mapping for `gc -> gc`, `gc -> value`, `value -> value`, `value -> gc` structs of the **same** type
* zero-initialisation for `gc -> gc`, `gc -> value`, `value -> value`, `value -> gc` structs of the **different** types

It also fixes a bug with "identity" memory mapping; i.e. when only an old type pointer needed to be replaced with a new pointer to the same type information, instead of actually mapping memory.


crates/mun_memory/src/lib.rs
```diff
@@ -24,6 +24,8 @@ pub trait TypeDesc: Send + Sync {
 pub trait TypeLayout: Send + Sync {
     /// Returns the memory layout of this type.
     fn layout(&self) -> Layout;
+    /// Returns the memory kind of this type, if it is a struct.
+    fn memory_kind(&self) -> Option<abi::StructMemoryKind>;
```
baszalmstra, line 28: Would it be possible to change this to:
`fn is_stack_allocated()` or something like that to avoid a dependency on the abi?

Wodann: Sounds good. Will fix that.

Wodann:
I've added a commit that resolves your comment. Could you please review it @baszalmstra?

PR #192, opened 2020-05-16
fix: remove return-blocking semicolon from ch01 listing 01

RadicalZephyr:
Hi there, I was reading through the Mun book, and noticed that this example doesn't compile as written in the book.


baszalmstra:
thanks!

PR #194, opened 2020-05-16
Fix code blocks in ch02-02-functions of the book

benediktwerner:
Markdown doesn't close code-blocks when the closing ```` ``` ```` isn't on a separate line.

[Rendering the change](https://github.com/mun-lang/mun/pull/194/files?short_path=6df04f5#diff-6df04f553edc809e85cc5b754ca0211b) shows the difference and it also happens [in the rendered book](https://docs.mun-lang.org/v0.2/ch02-02-functions.html#function-access-modifier).


baszalmstra:
Thanks very much! 👍

PR #197, opened 2020-05-17
Fix missing arg for fibonacci fn

fominok:
We need to pass `arg`


baszalmstra:
Good find! Thanks for the fix!

Issue #151, opened 2020-05-01
Update to latest LLVM


baszalmstra:
Currently, Mun uses LLVM 7.x.

It is beneficial for us to update to the latest LLVM but this requires some work to be compatible with the latest Inkwell.

Wodann:
@joshuawarner32 has commenced work on this

PR #201, opened 2020-05-23
Incremental codegen refactor to support llvm/inkwell upgrade

joshuawarner32:
This is a WIP refactor of the codegen, with the end goal of allowing upgrading llvm to 10 and inkwell to the upstream version (#151). Critically in the upstream version of inkwell many of the types grow a <'ctx> type parameter (For the lifetime of the inkwell::context::Context object), which is not compatible with how we are using Salsa in mun_codegen.

To that end, this is a refactoring that removes the use of Salsa from the mun_codegen. IrDatabaseStorage is replaced with four separate items:
* context: &Context (the inkwell context)
* type_manager: &mut TypeManager (caches struct/type info like salsa used to)
* config: &CodeGenConfig (holding the opt level and TargetData)
* db: &impl hir::HirDatabase (the salsa context from mun_hir)

This massively increases the length of many method signatures, which I'm not a huge fan of - but I suspect some of these will need careful lifetime annotations, which might be harder to get right if we bundle everything together prematurely. For sure context will need to be passed around separately.

(Note, the first two commits here are 100% temporary; I need to rebase these away before this is merged)


baszalmstra:
Hey @joshuawarner32 !

After some further investigation, we spotted some issues with our current design in Mun regarding storing the inkwell Context in a salsa Database. To do parallelism we would need to create a `Context` per thread. This conflicts with the current design that a `Context` is stored in a salsa database that is shared between threads. We discussed how we're going to solve this and decided that for now we're just not gonna use Salsa for the code generation. This is basically what you were doing all along 😅 . 

I just wanted to let you know that we're currently taking that route. I was looking at your PR to see if we could salvage some of what you already made, but your PR is way behind the current `master` branch. 

I'll be working on getting this up and running in the next couple of days, if you would still like to help out, that'd be more than welcome!

joshuawarner32:
Oh interesting!

> To do parallelism we would need to create a Context per thread.

My immediate reaction is that this could make the lifetimes very tricky, since all the different contexts would technically have different lifetimes. But maybe this is ok, depending on at what level you do the multithreading at. You probably can't have objects tied to one thread be accessed on another. (??)

PR #202, opened 2020-05-23
refactor(codegen): type safe inkwell types

baszalmstra:
Inkwell uses types that lose a lot of type information. An `u8`, `u16`, `i8`, and `16` are all represented by an `IntValue`. This takes some careful management of these types. For `IntValue` the problem is not so bad but for `PointerValue` and `StructValue` this is a real issue. By using `StructValue` you lose all information in the type with what kind of value you are actually working. This is not so much a problem for dynamic types (like Mun structs) but it is a major risk when using _abi_ types. I feel that this is major technical depth. 

This PR introduces a new type `Value<T>` that contains an inkwell value but it is typed on `T` which is the type that constructed the value. This allows you to write:

```rust
let a:Value<SomeStruct>;           // Holds an inkwell StructValue
let b:Value<*const SomeStruct>;    // Holds an inkwell PointerValue
```

It also allows much cleaner conversion from regular Rust types to inkwell types. e.g.

```rust
let a:BasicValueEnum = 3u32.as_value(context).into();
let b:ArrayValue = [0.0f32, 6.0f32].as_value(context).into();
let c:StructValue = (90, true).as_value(context).into();
```

There are 3 ways to enable custom types to work with this system:

1. Implement the `TransparentValue` trait which converts the implementor into another `Value<T>`. But in turn it also allows the implementor to be represented as a `Value<Self>`:
   ```rust
   use mun_codegen::value::{AsValue, IrValueContext, TransparentValue, Value};
   struct Foo {
      value: u32,
      bar: f32,
   }
   
    impl TransparentValue for Foo {
       type Target = (u32, f32);   

        fn as_target_value(&self, c ontext: &IrValueContext) -> Value<Self::Target> {
           (self.value, self.bar).as_value(context)
       }
   }
   ```
   The resulting IR type will be an anonymous type: ```type { u32, f32 }```

2. Auto derive the `AsValue` trait e.g.:

    ```rust
    #[derive(AsValue)]
    #[ir_name = "Foo"]
    pub struct Foo {
        value: u32,
        bar: f32,
    }
    ```

   The resulting IR type will be a named type: ```%Foo = type { u32, f32 }```

3. You can also implement all the support traits yourself for a custom type. 

I already found a few bugs in the LLVM ir that have been fixed.

### TODO
- [x] Refactor `abi_types.rs` to use the types from `Value<T>`s
- [x] Refactor the symbol and type-table generation to use the safe new `Value<T>` types
- [x] Create a procedural macro to validate that a certain structs matches its abi counterpart. (`ir::TypeInfo` should equal `abi::TypeInfo`)
- [x] See if we can remove some of the previous implementations (in `ir.rs`) (do in a later PR)


crates/mun_codegen/src/lib.rs
```diff
@@ -4,15 +4,20 @@ mod db;
 #[macro_use]
 mod ir;
 
+#[macro_use]
+extern crate mun_codegen_macros;
```
Wodann, line 8: Rust has already removed the need for this syntax, now allowing you to use syntax similar to `use some_module::some_fn;` in the file where you use the macro.

crates/mun_codegen/src/snapshots/test__arithmetic_op_f32.snap
```diff
@@ -6,9 +6,9 @@ expression: "pub fn add(a: f32, b: f32) -> f32 { a + b }\npub fn subtract(a: f32
 ; ModuleID = 'main.mun'
 source_filename = "main.mun"
 
-%struct.MunTypeInfo = type { [16 x i8], i8 addrspace(4)*, i32, i8, i8 }
+%struct.MunTypeInfo = type { [16 x i8], i8*, i32, i8, i8 }
 
-@global_type_table = external global [1 x %struct.MunTypeInfo addrspace(4)*]
+@global_type_table = external global [1 x %struct.MunTypeInfo*]
```
Wodann, line 11: I noticed that
```
@global_type_table = external global [1 x %struct.MunTypeInfo*]
```
and
```
@global_type_table = constant [1 x %struct.MunTypeInfo*] [%struct.MunTypeInfo* @"type_info::<core::i64>"]
```
now differ between `global` and `constant`. Why is that?

P.S. This applies to all snapshots

crates/mun_codegen/src/ir/file_group.rs
```diff
@@ -75,14 +73,18 @@ pub(crate) fn ir_query(db: &impl IrDatabase, file_id: hir::FileId) -> Arc<FileGr
...
-        db,
-        &llvm_module,
-        &abi_types,
-        intrinsics_map.keys(),
-        &dispatch_table,
-    );
+    let type_context = IrTypeContext {
+        context: &db.context(),
+        target_data: &db.target_data(),
+        struct_types: Default::default(),
```
Wodann, line 79: Was there a particular reason that you chose not to create a "helper" construction function? Is there a case where you want to start with some `struct_types` for example?

crates/mun_codegen/src/ir/types.rs
```diff
@@ -0,0 +1,113 @@
...
+#[derive(AsValue, TestIsAbiCompatible)]
+#[ir_name = "struct.MunAssemblyInfo"]
+#[abi_type(abi::AssemblyInfo)]
+pub struct AssemblyInfo {
+    pub symbols: ModuleInfo,
+    pub dispatch_table: DispatchTable,
+    pub dependencies: Value<*const *const u8>,
+    pub num_dependencies: u32,
+}
+
```
Wodann, line 111: cool stuff 👏

crates/mun_codegen/src/ir/types/test.rs
```diff
@@ -0,0 +1,192 @@
...
+/// self::test::AbiTypeHelper::from_value(&abi_value.<field_name>)
+///     .ir_type::<<field_type>>()
+///     .assert_compatible(<struct_type_name>, <abi_type_name>, <name>);
+/// ```
+pub trait TestIsAbiCompatible<T> {
+    /// Runs a test to see if the implementor is compatible with the specified abi type.
+    fn test(abi_value: &T);
+}
+
+/// A trait that is implemented if a type is ABI compatible with with T.
```
Wodann, line 25:
```suggestion
/// A trait that is implemented if a type is ABI compatible with `T`.
```

crates/mun_codegen/src/ir/types/test.rs
```diff
@@ -0,0 +1,192 @@
...
+    fn assert_compatible(&self, ir_type: &str, abi_type: &str, field_name: &str) {
+        panic!(
+            "the field '{}' on type '{}' is not compatible with the ABI version of the struct ({})",
+            field_name, ir_type, abi_type
+        );
+    }
+}
+impl<T> IsNotAbiCompatible for T {}
+
+/// Helper structs that enables extracting the type of a value.
```
Wodann, line 40:
```suggestion
/// Helper structs that allow extracting the type of a value.
```

crates/mun_codegen/src/ir/types/test.rs
```diff
@@ -0,0 +1,192 @@
...
+impl IsAbiCompatible<u16> for u16 {}
+impl IsAbiCompatible<u32> for u32 {}
+impl IsAbiCompatible<abi::Guid> for abi::Guid {}
+impl IsAbiCompatible<abi::TypeGroup> for abi::TypeGroup {}
+impl IsAbiCompatible<abi::StructMemoryKind> for abi::StructMemoryKind {}
+impl IsAbiCompatible<*const ::std::os::raw::c_char> for *const u8 {}
+impl IsAbiCompatible<*const ::std::os::raw::c_void> for *const fn() {}
+impl<S, T: IsAbiCompatible<S>> IsAbiCompatible<*const S> for *const T {}
+impl<S, T: IsAbiCompatible<S>> IsAbiCompatible<*mut S> for *mut T {}
+impl<S, T: ConcreteValueType> IsAbiCompatible<S> for Value<T> where T: IsAbiCompatible<S> {}
```
Wodann, line 68: Sweet! 💯

crates/mun_codegen/src/value/global.rs
```diff
@@ -0,0 +1,144 @@
...
+    values::{BasicValueEnum, PointerValue, UnnamedAddress},
+    AddressSpace,
+};
+use std::marker::PhantomData;
+
+/// Represents a typed global value. A `Global<T>` can be constructed from any `Value<T>` that can
+/// be converted to a [`inkwell::values::BasicValueEnum`].
+///
+/// Globals can be used to store data inside an inkwell context which can be referenced to from
+/// code.
```
Wodann, line 18:
```suggestion
/// Globals can be used to store data inside an inkwell context which can be referenced from code.
```

crates/mun_codegen/src/value/global.rs
```diff
@@ -0,0 +1,144 @@
...
+            value: self.value,
+            data: self.data,
+        }
+    }
+}
+
+impl<T: ?Sized> Copy for Global<T> {}
+
+impl<T: ?Sized> Global<T> {
+    /// Creates a `Global<T>` from an underlying value. Use with caution.
```
Wodann, line 38: If `use with caution`, should it be unsafe?

crates/mun_codegen/src/value/global.rs
```diff
@@ -0,0 +1,144 @@
...
+            global.set_unnamed_address(addr);
+        }
+        Global {
+            value: global,
+            data: Default::default(),
+        }
+    }
+
+    /// Converts self into a private const global. A private const global always has Private linkage
+    /// so its only accessible from the module its defined in. Its address is globally insignificant
```
Wodann, line 110:
```suggestion
    /// so its only accessible from the module it's defined in. Its address is globally insignificant
```

crates/mun_codegen/src/value/array_value.rs
```diff
@@ -0,0 +1,196 @@
...
+                    Value::from_raw(value)
+                }
+            }
+        )+
+    }
+);
+
+impl_array!(
+    0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 20, 24, 32, 36, 0x40, 0x80, 0x100
+);
```
Wodann, line 76: Does this imply that we only support arrays of particular sizes? If so, should we restrict it to the sizes that we actually need right now, and if someone introduces a new array size throw a `panic!` that directs to this code line?

crates/mun_codegen/src/value/string.rs
```diff
@@ -0,0 +1,56 @@
...
+    fn as_target_value(&self, context: &IrValueContext) -> Value<Self::Target> {
+        self.as_bytes_with_nul().as_value(context)
+    }
+}
+
+impl TransparentValue for String {
+    type Target = [u8];
+
+    fn as_target_value(&self, context: &IrValueContext) -> Value<Self::Target> {
+        self.as_bytes().as_value(context)
```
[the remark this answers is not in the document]
baszalmstra: No, not at all. Why do you ask? 

For clarity. I added both conversions from CString and String because they have different behaviors (can contain 0's, encoding, and null terminator). In the ABI we have to use CStrings so there we use that. If however, we added a string type later on I imagine we want to stay close to rusts String type.

crates/mun_codegen/src/ir/file_group.rs
baszalmstra: Its because `IrTypeContext::new(&db.context(),..)` doesn't work because the value of `db.context()` is dropped immediately in the call. So I would have to store the context first (and the target_data) which made the code longer and less clear IMHO.

crates/mun_codegen/src/ir/type_table.rs
```diff
@@ -190,8 +193,16 @@ impl<'a, 'ctx, 'm, D: IrDatabase> TypeTableBuilder<'a, 'ctx, 'm, D> {
                     self.value_context,
                 )
                 .as_value(self.value_context),
-            size_in_bits: type_info.size.bit_size as u32,
-            alignment: type_info.size.alignment as u8,
+            size_in_bits: type_info
+                .size
+                .bit_size
+                .try_into()
+                .expect("could not convert size in bites to smaller size"),
```
Wodann, line 200:
```suggestion
                .expect("could not convert size in bits to smaller size"),
```

crates/mun_codegen/src/value/array_value.rs
Wodann: Okay, the compile error is a generic error highlighting that this type is not implemented, right? Is it clear from the the compile error that someone can resolve it by editing this `impl_array!` invocation?

crates/mun_codegen/src/snapshots/test__arithmetic_op_f32.snap
Wodann: Thanks for the explanation.

crates/mun_codegen/src/value/string.rs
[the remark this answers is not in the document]
Wodann: I was thinking that our safe Inkwell wrapper would only be applicable to symbol generation. As that has to conform to the C ABI, I thought using the `String` and `str` types directly would potentially break that.

crates/mun_codegen/src/value/array_value.rs
baszalmstra: No you get the error AsValue (or some other trait) is not implemented for [A; 17]. I don't know how we could improve that though??

Wodann: Hmm. I am not sure if this would work work, but I saw that you had a default trait implementation somewhere, while you also had concrete trait implementations for some specific types. Would it be possible to have a default trait implementation for all array types that has does a [`compile_error`](https://doc.rust-lang.org/std/macro.compile_error.html)?

baszalmstra: No because every sized array type has a different signature. Thats the problem in the first place. Rust doesnt have generic consts (yet).

Wodann: Okay. I guess just leave it as is for now then 🙂

Issue #203, opened 2020-05-23
Project Management 0.1


Wodann:
To support multi-file Mun libraries, we need to come up with a project management system that allows users to easily setup, build, and tweak packages.

A Mun project will look as follows:
![image](https://user-images.githubusercontent.com/6917585/82733212-9f8cd300-9d12-11ea-9df5-579d27e7ba88.png)

The main directory contains a `src` directory and `mun.toml` configuration file. All files in the `src` directory are automatically added to the build tree. This is different from Rust in that you cannot/do not have to write `mod a` in your source code.

A sub-module can be created by either:
- adding a directory `a` with a `mod.mun` file; or
- adding a `a.mun` file in the parent directory.

> The above screenshot shows both of these examples for illustrative purposes.

To use a struct `Foo`, defined in `src/a/b.mun`, you can access it in the current file by writing: `use a::b::Foo;`.

By default, a type or function defined in a module is not accessible outside of its file and sub-modules. You can expand accessibility in three ways:

- `pub`: accessible within the package and externally (incl. marshalling to the host language)
- `pub(package)`: accessible within the package
- `pub(super)`: accessible within parent module and its sub-modules

A `mun.toml` contains information about the package, similar to a `Cargo.toml`:

```toml
[package]
name="test"
version="0.2.0"
authors = ["Mun Team"]
```

**Depends on**
- [x] #248 
- [x] #249

**Improved by**
- [x] #145
- [x] #168

**Remaining issues moved to later milestones**
- #425

nilq:
Looks reasonable. I like this structure; it provides  a nice mix between especially Rust's modules, but also Lua's module system. Would the `mun.toml` file be required for all Mun-libs?

baszalmstra:
The goal was to create something that is a bit simpler than Rusts method but without losing the ability to seperate functionalities. The design is also inspired by Typescript/Javascript.

I think the mun.toml file will be required yes. WDYT @Wodann ?

Wodann:
I don't see how we could not have a config file. We could do without the currently listed items for sure, but going forward we'll want to add settings such as:

- Hot reloading granularity: per-file, per-module, per-package
- Dependencies

Those will require a `mun.toml` file for sure.

Wodann:
To add to my previous comment, we want to have full compile flag support alongside the `mun.toml` file. For those people who don't want to use our project manager, they can use compiler flags instead.

nilq:
[...]

PR #223, opened 2020-06-12
feature: foundation for projects

baszalmstra:
This PR is an initial implementation of #203. It introduces the concept of a package as well as replacing the build step from the `mun` executable with a build step that expects a `mun.toml` to be present in the working directory (or one of its parents), similar to how `cargo build` works.  The goal of this PR is just to provide the foundation that we can build on in later PRs.

#### Packages

The contents of the `mun.toml` should be similar to:

```
[package]
name="foobar"
author=["John Doe"]
version="0.1.0"
```

The directory structure of a project should look like this

```
package_dir
|- src
|   \- main.mun
\- mun.toml
```

#### Multiple modules?!

This PR also adds multifile support, you can have multiple files in the source directory however there is no linking between them whatsoever, they are currently treated as separate libraries. (see below)

#### Whats missing?

There are a bunch of things missing from this PR that should be implemented at a later stage:

- [ ] Update the docs (because a project is now required)
- [ ] We should add a `mun init`/`mun new` command to initialize an empty package. (similar to cargo)
- [ ] Modules should be able to interact with each other and be compiled to a single package (depending on hotreload settings).
- [ ] Add proper integration tests where we actually test:
   1. create a project
   2. run `mun build --watch` (or without `--watch`)
   3. load the binary and invoke it (testing its functionality)
   4. modify the source
   5. test the functionality again


crates/mun_project/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.1.0"
+authors = ["The Mun Team <team@mun-lang.org>"]
+edition = "2018"
+
+[dependencies]
+serde="1.0"
+serde_derive="1.0"
+toml="0.5"
+semver={ version = "0.10", features=["serde"] }
+anyhow="1.0"
```
Wodann, line 12:
```suggestion
serde = "1.0"
serde_derive = "1.0"
toml = "0.5"
semver = { version = "0.10", features = ["serde"] }
anyhow = "1.0"
```

crates/mun_compiler/src/driver.rs
```diff
@@ -77,15 +86,101 @@ impl Driver {
...
+        relative_path: P,
+    ) -> Result<FileId, anyhow::Error> {
+        // Re-use existing id to get better caching performance
+        if let Some(id) = self.path_to_file_id.get(relative_path.as_ref()) {
+            return Ok(*id);
+        }
+
+        // Allocate a new id
+        // TODO: See if we can figure out if the compiler cleared the cache of a certain file, at
+        //  which point we can sort of reset the `next_file_id`
```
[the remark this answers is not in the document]
baszalmstra: This is correct because "which" belongs to the previous line (required because otherwise the IDE will not see the entire string as a TODO (only the first part)).

baszalmstra:
> Additionally, not writing tests moves the workload to the next person to touch this, which is not fair. It's our policy to ask contributors to add sufficient tests to their PRs, so we should do so ourselves too 🙂
Yes you are right. Do you suggest I add the test flow I described above?

Wodann:
> Yes you are right. Do you suggest I add the test flow I described above?

There are two approaches you can take. I think a lot of the functions you created actually lend themselves to unit tests. That would already bring a major improvement, I think.

For the parts that can only be tested through integration tests, your proposed scheme looks good. The major functionalities for integration tests - which I can think of right now - are:
- create file
- edit file
- rename file
- delete file

baszalmstra:
Ive added some more tests. The `build` part of the executable is now tested by compiling and running a simple mun project. Codecov doesnt report this correctly though..

The crates `mun_project` currently has 90%+ coverage.

crates/mun_runtime/src/lib.rs
```diff
@@ -3,6 +3,7 @@
 //! The Mun Runtime provides functionality for automatically hot reloading Mun C ABI
 //! compliant shared libraries.
 #![warn(missing_docs)]
+#![allow(clippy::many_single_char_names)]
```
Wodann, line 6: Is this still necessary?

crates/mun_runtime/src/macros.rs
```diff
@@ -89,7 +89,7 @@ macro_rules! invoke_fn_impl {
                 ///
                 /// If an error occurs when invoking the method, an error message is logged. The
                 /// runtime continues looping until the cause of the error has been resolved.
-                #[allow(clippy::too_many_arguments, unused_assignments)]
+                #[allow(clippy::too_many_arguments, unused_assignments, clippy::many_single_char_names)]
```
Wodann, line 92: Is this still necessary?

crates/mun_runtime/src/lib.rs
baszalmstra: Nope removed!

crates/mun_runtime/src/macros.rs
baszalmstra: Nope removed!

baszalmstra:
Relates to #203

PR #226, opened 2020-06-15
improvement: added combatibility check between munlib and runtime

### LINKED ISSUE #188: Implement compatibility checks between munlib and runtime
**Problem**

Currently, the runtime tries to load a `*.munlib` with the assumption that the ABI memory layout of the `*.munlib` matches its own ABI version. This can cause undefined behaviour when it tries to load a `*.munlib` with a different ABI version.

**Solution**

By including an ABI version in the `*.munlib`, the runtime can gracefully fail - reporting an error to the end user.

In the future - once we support backwards compatibility, we can even support loading of several different ABI versions using this functionality.

**Tasks**

- [ ] add an integer `ABI_VERSION` constant to the `mun_abi` crate.
- [ ] [generate](https://github.com/mun-lang/mun/blob/0d6e36d64376289a8969c62cefdc080ee99c2a51/crates/mun_codegen/src/code_gen/symbols.rs#L291) a `get_version` IR function
- [ ] [export](https://github.com/mun-lang/mun/blob/c17448d77f9751e74ce90ea49cd8503b35459d8e/crates/mun_codegen/src/code_gen/linker.rs#L176) the `get_version` function
- [ ] in the [runtime](https://github.com/mun-lang/mun/blob/master/crates/mun_runtime/src/assembly.rs#L35), call the `get_version` function and verify that the runtime's `mun_abi::ABI_VERSION` is compatible.
- [ ] (bonus) add a constant with the `get_version` function name (and other API functions) to the `mun_abi` crate and use that constant in both the `mun_codegen` and `mun_runtime` crates - to avoid discrepancies.

**Good first issue process:**

- [ ] implement the feature and start a Pull Request with in the description: closes #188. 

MineBill:
This PR closes #188

### DIFF
--- a/crates/mun_abi/src/lib.rs
+++ b/crates/mun_abi/src/lib.rs
@@ -26,7 +26,8 @@ pub mod prelude {
 }
 
 /// Defines the current abi version
-pub const ABI_VERSION: i32 = 00_20_00;
+#[allow(clippy::zero_prefixed_literal)]
+pub const ABI_VERSION: u32 = 00_02_00;
 /// Defines the name for the `get_info` function
 pub const GET_INFO_FN_NAME: &str = "get_info";
 /// Defines the name for the `get_version` function
--- a/crates/mun_abi/src/lib.rs
+++ b/crates/mun_abi/src/lib.rs
@@ -25,6 +25,15 @@ pub mod prelude {
     };
 }
 
+/// Defines the current abi version
+pub const ABI_VERSION: i32 = 00_20_00;
+/// Defines the name for the `get_info` function
+pub const GET_INFO_FN_NAME: &str = "get_info";
+/// Defines the name for the `get_version` function
+pub const GET_VERSION_FN_NAME: &str = "get_version";
+/// Defines the name for the `set_allocator_handle` function
+pub const SET_ALLOCATOR_HANDLE_FN_NAME: &str = "set_allocator_handle";
+
 /// Represents the kind of memory management a struct uses.
 #[repr(u8)]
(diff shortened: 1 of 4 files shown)


crates/mun_abi/src/lib.rs
```diff
@@ -25,6 +25,15 @@ pub mod prelude {
     };
 }
 
+/// Defines the current abi version
+pub const ABI_VERSION: i32 = 00_20_00;
```
Wodann, line 29: Did you have a reason in mind for choosing a signed integer (`i32`) over an unsigned integer (`u32`)?

Wodann:
Nice work! 👏 

I have one [question](https://github.com/mun-lang/mun/pull/226/files#r440114416)

crates/mun_abi/src/lib.rs
MineBill: Not really, should i change it to a u32?

Wodann: I don't think we'll ever hit the maximum number of versions possible with `i32`, but still I think that `u32` would be more correct - as I don't think we'll ever use negative version numbers.

Wodann, line 29:
```suggestion
#[allow(clippy::zero-prefixed-literal)]
pub const ABI_VERSION: i32 = 00_02_00;
```

Wodann: I am not sure whether @baszalmstra agrees, but I think we can ignore this clippy warning. I like the general formatting of the version number into major, minor, patch.

I did change minor `20` to `02`

PR #232, opened 2020-06-19
fix: fixes issues #225

### LINKED ISSUE #225: Compiler panics when accessing a field of a temporary
Trying to compile this:

```
struct Num {
    value: i64,
}

pub fn foo() {
    Num { value: 0 }.value;
}
```

results in

```
thread 'main' panicked at 'internal error: entered unreachable code: invalid place expression', <::std::macros::panic macros>:5:6
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

baszalmstra:
Fixes an issue where taking the field of a temporary crashed the compiler. This was caused by a field expression always expecting a "place" expression which is actually not always the case (in the case of a temporary for example).

Closes #225

### DIFF
--- a/crates/mun_codegen/src/ir/body.rs
+++ b/crates/mun_codegen/src/ir/body.rs
@@ -261,11 +261,7 @@ impl<'a, 'b, D: IrDatabase> BodyIrGenerator<'a, 'b, D> {
             Expr::Field {
                 expr: receiver_expr,
                 name,
-            } => {
-                let ptr = self.gen_field(expr, *receiver_expr, name);
-                let value = self.builder.build_load(ptr, &name.to_string());
-                Some(value)
-            }
+            } => self.gen_field(expr, *receiver_expr, name),
             _ => unimplemented!("unimplemented expr type {:?}", &body[expr]),
         }
     }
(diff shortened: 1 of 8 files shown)


crates/mun_codegen/src/ir/body.rs
```diff
@@ -569,18 +565,21 @@ impl<'a, 'b, D: IrDatabase> BodyIrGenerator<'a, 'b, D> {
...
                 ctor: hir::TypeCtor::Struct(s),
                 ..
             }) => match s.data(self.db).memory_kind {
-                hir::StructMemoryKind::GC => deref_heap_value(&self.builder, value),
-                hir::StructMemoryKind::Value => value,
+                hir::StructMemoryKind::GC => return deref_heap_value(&self.builder, value),
+                hir::StructMemoryKind::Value => {}
             },
-            _ => value,
+            _ => {}
```
Wodann, line 580: what is the actual change here?

baszalmstra: First it returned values from the match, but I found it clearer to have it fall through if it didn't match so the default behavior would be more obvious (simply return the value).

Wodann: Maybe turn this into an `if let` statements in that case? That would remove quite some loc and shows that point a little better imo

crates/mun_codegen/src/snapshots/test__issue_225.snap
```diff
@@ -0,0 +1,75 @@
...
+  %init = insertvalue %Num undef, i64 %0, 0
+  %new_ptr = load i8** (i8*, i8*)*, i8** (i8*, i8*)** getelementptr inbounds (%DispatchTable, %DispatchTable* @dispatchTable, i32 0, i32 0)
+  %Num_ptr = load %struct.MunTypeInfo*, %struct.MunTypeInfo** getelementptr inbounds ([5 x %struct.MunTypeInfo*], [5 x %struct.MunTypeInfo*]* @global_type_table, i32 0, i32 2)
+  %type_info_ptr_to_i8_ptr = bitcast %struct.MunTypeInfo* %Num_ptr to i8*
+  %allocator_handle = load i8*, i8** @allocatorHandle
+  %new = call i8** %new_ptr(i8* %type_info_ptr_to_i8_ptr, i8* %allocator_handle)
+  %Num_ptr_ptr = bitcast i8** %new to %Num**
+  %Num_mem_ptr = load %Num*, %Num** %Num_ptr_ptr
+  store %Num %init, %Num* %Num_mem_ptr
+  %mem_ptr = load %Num*, %Num** %Num_ptr_ptr
```
Wodann, line 28: **Future Work**
This is a great example of where we can optimise a lot by using a `struct(value)` instead of a `struct(gc)`. For temporaries (or scoped `struct` instances) it doesn't  really make sense to allocate them on the heap.

baszalmstra: That would be great! But the use case is quite small, it's only when you construct a GC struct that doesn't leave the scope of the function at all (otherwise you can't make guarantees). Nonethess, if we can catch that in the compiler and turn it into a stack allocation, that would be great!

PR #233, opened 2020-06-20
fix: fixes a lot of llvm assertions

### LINKED ISSUE #227: Generated llvm ir triggers llvm assertions
So I accidentally compiled my llvm with assertions enabled, and now trying to run mun tests aborts with llvm assertion failure:

```
>  cargo test -p mun_codegen -- function_arguments
    Finished test [unoptimized + debuginfo] target(s) in 0.34s
     Running target\debug\deps\mun_codegen-d51557810a1e1457.exe

running 1 test
Assertion failed: V[I]->getType() == T->getTypeAtIndex(I) && "Initializer for composite element doesn't match!", file ###\lib\IR\Constants.cpp, line 951
error: test failed, to rerun pass '-p mun_codegen --lib'

Caused by:
  process didn't exit successfully: `###\target\debug\deps\mun_codegen-d51557810a1e1457.exe function_arguments` (exit code: 0xc0000409, STATUS_STACK_BUFFER_OVERRUN)
```

Looking at the [ir snapshot](https://github.com/mun-lang/mun/blob/ec0c49ddbd70acf14cf9c8ce49d9709bbd46d524/crates/mun_codegen/src/snapshots/test__function_arguments.snap) my suspicion is that type info initializer does not match the declared type `MunTypeInfo`:

```
%struct.MunTypeInfo = type { [16 x i8], i8*, i32, i8, i8 }
                                        ^^^ declared as i8*
...

@"type_info::<core::i32>" = private unnamed_addr constant %struct.MunTypeInfo { [16 x i8] c"\17yzt\19\D62\17\D25\95C\17\88[\FA", [10 x i8]* @"type_info::<core::i32>::name", i32 32, i8 4, i8 0 }
                                                                                                                                 ^^^^^^^^^^ initializer uses [10 x i8]*

```

baszalmstra:
Fixes a lot of LLVM assertions that were caused by invalid struct assignments due to casts being invalid. I introduced the `Value::with_cast` method to introduce an explicit cast.

Fixes these issues was a pain because the test just crashed. I had to insert a lot of print statements. I left some of them in (commented out) so in the future, it will be easier to fix these issues.

Closes #227


crates/mun_codegen/src/snapshots/test__issue_225.snap
```diff
@@ -0,0 +1,75 @@
+---
+source: crates/mun_codegen/src/test.rs
+expression: "struct Num {\n    value: i64,\n}\n\npub fn foo(b: i64) {\n    Num { value: b }.value;\n}\n\npub fn bar(b: i64) {\n    { let a = Num { value: b }; a}.value;\n}"
```
Wodann, line 3: This is from another PR, so maybe merge that one first and then rebase this PR

crates/mun_codegen/src/snapshots/test__main.snap
```diff
@@ -0,0 +1,58 @@
+---
+source: crates/mun_codegen/src/test.rs
```
Wodann, line 2: This test name seems incorrect for the included tests. Was this added by mistake?

crates/mun_codegen/src/snapshots/test__issue_225.snap
baszalmstra: Done

PR #236, opened 2020-06-27
feature: initial LSP support

tdejager:
This MR adds support for the initial LSP scaffolding and can be started from the **mun** main executable using:

```bash
mun language-server
```

Note that currently only diagnostics are supported, but it is a nice starting point for future developments. The README now contains instructions on how to use this in either `vscode` or `vim`.

PR #239, opened 2020-07-04
Feature: shared diagnostics between compiler and language server

baszalmstra:
This PR introduces the `mun_diagnostics` crate. This crate provides the `Diagnostic` trait which is used by both the compiler and the language server to provide the same detailed diagnotics. The crate enables conversion from `hir::Diagnostic` to `mun_diagnostics::Diagnostic` which also adds more detailed information (which first resided in the compiler itself).

The compiler uses the `Diagnostic` trait to convert it to a `Snippet` from annotate_snippets to emit the diagnostic information to the console.

The language server uses the `Diagnostic` trait to convert it to a `lsp_types::Diagnostic` which is send to the language server client. 

This enables sharing a lot of the logic of diagnostics between these two use cases.

In the future, we could also add quick-fixes to the crate to be able to provide hints on the console and provide quick actions for the LSP.

This PR is branched of from #236 which should be merged first.


crates/mun_diagnostics/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+        Vec::new()
+    }
+
+    /// Optional footer text
+    fn footer(&self) -> Vec<String> {
+        Vec::new()
+    }
+}
+
+pub trait DiagnosticFor {
```
Wodann, line 55: Missing trait-level docs

crates/mun_diagnostics/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+    /// Calls the specified function `f` with an instance of a [`Diagnostic`]. This can be used
+    /// to perform lazy diagnostic evaluation.
+    fn with_diagnostic<R, F: FnMut(&dyn Diagnostic) -> R>(&self, f: F) -> R;
+}
+
+pub trait DiagnosticForWith<With> {
```
Wodann, line 61: Missing trait-level docs

crates/mun_hir/src/ty.rs
```diff
@@ -177,6 +177,14 @@ impl Ty {
             _ => None,
         })
     }
+
+    /// Returns true if the specified value is a valid type or whether an error occured with it.
```
Wodann, line 181: This is a little vague. Maybe rephrase it?

PR #241, opened 2020-07-08
Explicitly specify latest Ubuntu LTS

natrys:


natrys:
Github action, for whatever reason, still has ubuntu-latest pointed to ubuntu-18.04. Would you be interested having it explicitly pointed it to 20.04? The release binaries currently don't work in newer linux systems (due to being linked against old version of libffi).

Wodann:
Thank you for the PR! :slight_smile: Even though this change looks good, there is some concern about whether it will raise the same errors for older versions of Ubuntu.

The core team is looking into alternative fixes for the problem you are experiencing, to prevent it from regressing the situation for others. Please bare with us.

natrys:
Alright. Perhaps musl based static linking could be explored instead, if viable.

Wodann:
Thanks. I actually wasn't aware of that. I'll research musl as an option! 👍

PR #243, opened 2020-07-15
fix(parser): parsing of unmatched right curly braces ('}')

emigr2k1:


baszalmstra:
Thanks for this! 👍

Wodann:
Thanks for finding and fixing this bug @emigr2k1.

A recommendation for the future would be to also add a test that causes the crash to our test suite - along with the fix. That way we can check for regressions, going forward :)

PR #260, opened 2020-08-23
refactor: use custom prebuild llvm distribution on ubuntu

baszalmstra:


baszalmstra:
Closing this, we are going to keep it this way. But modify the artifact stage instead to build in a docker container that does static linking.

PR #266, opened 2020-09-11
refactor: use Idx<T> instead of macro in arena

baszalmstra:
This PR refactors the Arena Id's a little bit so that we need fewer macros and get a clearer coupling between Arenas, their data and their ids. Similar to how rust analyzer refactored this. 

I also renamed a few Ids to indicate that they are "Local" to their container (`TypeRefId` is local to a specific `TypeRefMap` and is now called `LocalTypRefId`). I found that to be clearer.

### DIFF
--- a/crates/mun_codegen/src/ir/body.rs
+++ b/crates/mun_codegen/src/ir/body.rs
@@ -6,9 +6,9 @@ use crate::{
     value::Global,
 };
 use hir::{
-    ArenaId, ArithOp, BinaryOp, Body, CmpOp, Expr, ExprId, HirDatabase, HirDisplay,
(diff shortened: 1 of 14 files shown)


Wodann (approved):

Wodann:
My only concern is the reduced codecov, otherwise it looks good.

It seems that codecov already has some good recommendations for improving the coverage.

baszalmstra:
> 
> 
> My only concern is the reduced codecov, otherwise it looks good.
> 
> It seems that codecov already has some good recommendations for improving the coverage.

Oh I only just noticed you also commented. The code coverage of the overall project actually increases because the MR removes a lot of code. Other than that (besides the Arena code, which is tested) I only really renamed stuff so I didn't feel like adding more tests would be necessary.

I do see that there are a few functions in there that dont seem to be used at all.

PR #268, opened 2020-09-17
WIP: refactor: replace `TypeInfo` with `TypeRef` for runtime linking of types

Wodann:


baszalmstra:
What is the status of this MR? This would be very useful for array support.

PR #269, opened 2020-09-18
build: pin rust version

### LINKED ISSUE #247: Pin Rust version
Everytime Rust releases a new version our builds break. This happens because new lints are added for instance. We want to keep up to date with the latest Rust version but its anoying that if someone submits a PR CI might fail because a new Rust version was just released.

I propose we do something similar to `vector`, they store [a file](https://github.com/timberio/vector/blob/master/rust-toolchain) that stores which version of Rust CI should use. Its trivial to bump Mun to a new Rust version by creating a new PR that modifies the contents of that file together with fixing any issues that may be caused by the version bump.

To implement this we just need to modify our CI workflow to read the version to use from a file.

**Good first issue process:**

- [x] [read our contributing guidelines](https://github.com/mun-lang/mun/blob/master/CONTRIBUTING.md)
- [x] claim this issue (assign yourself or comment below)
- [ ] implement the feature and start a Pull Request with in the description: closes #247 . 

sinato:
This PR pins the version of the rust toolchain to be used in the build and CI of this project ( closes #247 ) .

- [x] Add `rust-toolchain` file to pin the version used for `rustup`
  - ref: https://github.com/rust-lang/rustup/blob/bde7f787d7/doc/src/overrides.md#the-toolchain-file
- [x] Modify CI / release workflow to read `rust-toolchain` file (just remove the `toolchain` input)
  - ref: https://github.com/actions-rs/toolchain/tree/69f9ca11d94b20d95d67be4902f28e68f089145c#the-toolchain-file


baszalmstra:
With these changes, I think you can also remove the nightly and beta CI stages. They dont make sense to me if you pin the Rust version anyway. WDYT? what does vector do?

.github/workflows/ci.yml
```diff
@@ -19,7 +19,6 @@ jobs:
         uses: actions-rs/toolchain@v1
```
Wodann, line 19: Are both this toolchain step and newly added one in line 50 necessary?

sinato:
> With these changes, I think you can also remove the nightly and beta CI stages. They dont make sense to me if you pin the Rust version anyway. WDYT? what does vector do?

As far as I understand, `vector` only use the pinned version and stable. I also think beta and nightly are no longer needed. 
Fixed: a1fedce

.github/workflows/ci.yml
sinato: I used those two steps to choose whether to input the toolchain (use the specified version or use the pinned version) depending on whether it is stable or not.
There may be a better way to do this that I just don't know about...... but it was removed by  a1fedce anyway.

Wodann (approved):

Wodann:
Just checking, but did you want to squash some of the commits or can I merge it as is? Either is fine by me 🙂

baszalmstra (approved):

sinato:
Thank you for the review! Can you merge them as is? I think each commit meaningful enough.

PR #271, opened 2020-09-20
feature: refactored RawItems into ItemTree

baszalmstra:
First refactor to remove RawItems and create better separation of concerns. ItemTree is a derived AST that only contains module items. All interning is now done as ItemTree ids instead of building directly on the AST. Hopefully, this creates a clearer separation between items and the AST.

When adding modules (which Im also working on) having this ItemTree is much clearer than using either RawItems or the code model or a strange combination. Im not sure yet how to handle the `TypeRefSourceMap`, Im looking into removing that too.

### DIFF
--- a/crates/mun_hir/src/item_tree/lower.rs
+++ b/crates/mun_hir/src/item_tree/lower.rs
@@ -1,0 +1,204 @@
+//! This module implements the logic to convert an AST to an `ItemTree`.
+
+use super::{
+    Field, Fields, Function, IdRange, ItemTree, ItemTreeData, ItemTreeNode, LocalItemTreeId,
+    ModItem, Struct, StructDefKind, TypeAlias,
+};
+use crate::{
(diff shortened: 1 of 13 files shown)


crates/mun_hir/src/item_tree/lower.rs
```diff
@@ -0,0 +1,204 @@
...
+}
+
+pub(super) struct Context {
+    file: FileId,
+    source_ast_id_map: Arc<AstIdMap>,
+    data: ItemTreeData,
+}
+
+impl Context {
+    /// Constructs a new `LowerContext` for the specified file
```
Wodann, line 36:
```suggestion
    /// Constructs a new `Context` for the specified file
```

Wodann (changes requested):
Nice refactor. I don't have anything to improve, except a typo in the comments.

Wodann (approved):

PR #272, opened 2020-09-22
feature: adds fixtures to support multiple files from string

baszalmstra:
Adds the ability to construct test fixtures that consist of multiple files from a single string. A fixture with multiple files looks like this:

```rust
Fixture::parse(
    r#"
    //- /foo.mun
    fn hello_world() {
    }

    //- /bar.mun
    fn baz() {
    }
"#
)
```

This is used by the MockDatabase to enable creating tests for multiple files.

### DIFF
--- a/crates/mun_hir/Cargo.toml
+++ b/crates/mun_hir/Cargo.toml
@@ -27,3 +27,4 @@ either = "1.5.3"
 [dev-dependencies]
 insta = "0.16"
 parking_lot = "0.10"
+mun_test = { version = "=0.1.0", path = "../mun_test" }
--- a/crates/mun_test/src/fixture.rs
+++ b/crates/mun_test/src/fixture.rs
@@ -1,0 +1,221 @@
+use compiler::RelativePathBuf;
+use itertools::Itertools;
+
+const DEFAULT_FILE_NAME: &str = "main.mun";
+const META_LINE: &str = "//-";
+
+/// A `Fixture` describes an single file in a project workspace. `Fixture`s can be parsed from a
+/// single string with the `parse` function. Using that function enables users to conveniently
+/// describe an entire workspace in a single string.
+#[derive(Debug, Eq, PartialEq)]
+pub struct Fixture {
+    /// The relative path of this file
+    pub relative_path: RelativePathBuf,
+
+    /// The text of the file
+    pub text: String,
+}
+
+impl Fixture {
+    /// Parses text which looks like this:
+    ///
+    /// ```not_rust
+    /// //- /foo.mun
+    /// fn hello_world() {
+    /// }
+    ///
+    /// //- /bar.mun
+    /// fn baz() {
+    /// }
+    /// ```
+    ///
+    /// into two separate `Fixture`s one with `relative_path` 'foo.mun' and one with 'bar.mun'.
+    pub fn parse(text: impl AsRef<str>) -> Vec<Fixture> {
+        let text = trim_raw_string_literal(text);
+        let mut result: Vec<Fixture> = Vec::new();
+
+        // If the text does not contain any meta tags, insert a default meta tag at the start.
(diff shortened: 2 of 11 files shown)


crates/mun_hir/Cargo.toml
```diff
@@ -23,7 +23,9 @@ relative-path = "1.2"
 ena = "0.14"
 drop_bomb = "0.1.4"
 either = "1.5.3"
+itertools = "0.9.0"
```
Wodann, line 26: Does this crate actually use itertools?

crates/mun_test/src/fixture.rs
```diff
@@ -0,0 +1,221 @@
+use compiler::RelativePathBuf;
+use itertools::Itertools;
+
+const DEFAULT_FILE_NAME: &str = "main.mun";
+const META_LINE: &str = "//-";
+
+/// A `Fixtures` describes an single file in a project workspace. `Fixture`s can be parsed from a
```
Wodann, line 7:
```suggestion
/// A `Fixture` describes a multiple files in a project workspace. `Fixture`s can be parsed from a
```

crates/mun_test/src/fixture.rs
```diff
@@ -0,0 +1,221 @@
...
+    /// //- /foo.mun
+    /// fn hello_world() {
+    /// }
+    ///
+    /// //- /bar.mun
+    /// fn baz() {
+    /// }
+    /// ```
+    ///
+    /// into two separate `Fixtures` one with `relative_path` 'foo.mun' and one with 'bar.mun'.
```
Wodann, line 32:
```suggestion
    /// into two separate `Fixture`s one with `relative_path` 'foo.mun' and one with 'bar.mun'.
```

crates/mun_test/src/fixture.rs
```diff
@@ -0,0 +1,221 @@
...
+                    relative_path: RelativePathBuf::from("bar.mun"),
+                    text: "fn baz() {\n}\n".to_owned()
+                }
+            ]
+        );
+    }
+
+    #[test]
+    #[should_panic]
+    fn incorrect_indented_fixture() {
```
Wodann, line 211:
```suggestion
    fn incorrectly_indented_fixture() {
```

Wodann (changes requested):
Other than some grammar mistakes, this looks great! 👍

crates/mun_hir/Cargo.toml
baszalmstra: Nope, will remove

crates/mun_test/src/fixture.rs
baszalmstra: No thats actually correct. The struct itself describes a single file in a project workspace. When calling parse a collection is returned.

Wodann: The fixtures (plural class name) part or the singular file? It seems like you define multiple Mun files in your fixture, not a single one?

baszalmstra: It should be `A Fixture describes a single file in a project workspace.` Does that make sense?

Wodann: I guess my understanding of what a fixture is, is wrong then. I thought that the following snippet contained one fixture, comprised of two files.

```rust
Fixture::parse(
    r#"
    //- /foo.mun
    fn hello_world() {
    }

    //- /bar.mun
    fn baz() {
    }
"#
)
```

baszalmstra: Parsing the given strings returns a `Vec<Fixture>`. One entry for `foo.mun` and one for `bar.mun`.

baszalmstra:
Can we merge this @Wodann ?

Wodann (approved):

PR #283, opened 2020-10-10
feature: adds modules and visibility

### LINKED ISSUE #248: Add language support for specifying `pub(...)` scope
By default, a type or function defined in a module is not accessible outside of its file and sub-modules. You can expand accessibility in three ways:

- `pub`: accessible within the package and externally (incl. marshalling to the host language)
- `pub(package)`: accessible within the package
- `pub(super)`: accessible within parent module and its sub-modules

For an example, see #249.

baszalmstra:
This (still way too large, Im sorry 😓 ) PR adds use of `pub` keywords and resolved visibility. It also introduces the concept of modules which are ordered in a hierarchical manner based on their location on disk. This is required to be able to specify visibility scopes. This PR also (again Im sorry for the size 😓 ) introduces the concept of definitions from all over the package in the form of PackageDefs. This can later be used to more easily implement `use` as well. 

All of this required quite the overhaul, especially because we have modules now instead of files.

Closes #248

This PR is still in Draft because I want to:

- [x] ~~add diagnostics for missing module files (`foo/` exists but no `mod.rs` or `foo.rs`).~~ We decided that that is actually not an error.
- [x] add more tests!
- [x] test visibility resolution
- [x] Actually check visibility in paths


crates/mun_test/src/driver.rs
```diff
@@ -43,8 +43,13 @@ impl CompileTestDriver {
                     .expect("compiler errors are not UTF-8 formatted")
             )
         }
-        let out_path = driver.assembly_output_path(file_id);
-        driver.write_assembly(file_id, true).unwrap();
+        //
```
Wodann, line 46: Can this be removed?

crates/mun_syntax/src/tests/parser.rs
```diff
@@ -365,3 +365,14 @@ fn type_alias_def() {
     "#,
     )
 }
+#[test]
+fn function_return_path() {
+    snapshot_test(
+        r#"
+        fn main() -> self::Foo {}
```
Wodann, line 372: Do we want to support `self` for paths? Based on previous discussion we had set these:
image

crates/mun_syntax/src/parsing/grammar/paths.rs
```diff
@@ -1,9 +1,10 @@
 use super::*;
 
-pub(super) const PATH_FIRST: TokenSet = token_set![IDENT, SELF_KW, SUPER_KW, COLONCOLON];
+pub(super) const PATH_FIRST: TokenSet =
+    TokenSet::new(&[IDENT, T![super], T![self], T![package], T![::]]);
 
 pub(super) fn is_path_start(p: &Parser) -> bool {
-    matches!(p.current(), IDENT | T![::])
+    matches!(p.current(), IDENT | T![self] | T![super] | T![package])
```
Wodann, line 7: Do we need `self`?

crates/mun_hir/src/ty/tests.rs
```diff
@@ -1,10 +1,161 @@
...
+    pub(super) struct PubSupBar(i64);
+    pub(super) struct PubSupBaz;
+    pub(super) fn pub_sup_foo() {}
+
+    pub(package) struct PubPackageFoo {};
+    pub(package) struct PubPackageBar(i64);
+    pub(package) struct PubPackageBaz;
+    pub(package) fn pub_package_foo() {}
+
+    //- /bar.mun
```
Wodann, line 33: Does it make sense to add test cases for type aliases?

crates/mun_hir/src/visibility.rs
```diff
@@ -0,0 +1,87 @@
+use crate::{ids::ModuleId, DefDatabase, HirDatabase, Resolver};
+use mun_syntax::ast;
+use std::iter::successors;
+
+/// Visibility of an item, not yet resolved to an actual module.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum RawVisibility {
+    /// Accessible from self (Self)
```
Wodann, line 8: Also sub-modules, right?

crates/mun_hir/src/visibility.rs
```diff
@@ -0,0 +1,87 @@
...
+    This,
+
+    /// Accessible from self or parent module
```
Wodann, line 11: Also sub-modules, right?

crates/mun_hir/src/visibility.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// Returns true if an item with this visibility is accessible from the given module.
+    pub fn is_visible_from(self, db: &dyn HirDatabase, from_module: ModuleId) -> bool {
+        let to_module = match self {
+            Visibility::Module(m) => m,
+            Visibility::Public => return true,
+        };
+
+        let module_tree = db.module_tree(from_module.package);
+        let mut ancestors = successors(Some(from_module.local_id), |m| {
+            let parent_id = module_tree[*m].parent?;
```
Wodann, line 69: if you are using the `?` operator here, can't you just directly return `module_tree[*m].parent`, without the `?` and `Some` constructor?

Wodann (changes requested):
Very happy to see this work nearing completion. Mun v0.3 is shaping up nicely!

This is only a partial review, but I wanted to make sure that part of my comments were already actionable. I'll continue the review later.

crates/mun_hir/src/code_model/module.rs
```diff
@@ -0,0 +1,108 @@
...
+        let package_defs = db.package_defs(self.id.package);
+        package_defs.modules[self.id.local_id]
+            .declarations()
+            .map(ModuleDef::from)
+            .collect()
+    }
+
+    /// Iterate over all diagnostics from this `Module` by placing them in the `sink`
+    pub fn diagnostics(self, db: &dyn HirDatabase, sink: &mut DiagnosticSink) {
+        if let Some(file_id) = self.file_id(db) {
```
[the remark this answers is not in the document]
baszalmstra: This is local to this module.

crates/mun_syntax/src/tests/parser.rs
baszalmstra: `pub(self)` is not valid. 

`self` can only be used in module paths. I use `self` sometimes to make it explicit that the given type or function comes from this module. In the test examples, it is also used to explicitly distinguish the `Foo` coming from this module. I also use it to reexport something from a child module: `pub use self::foo::Bar`

Do you think that the redundancy outweighs the use of the explicitly distinguishing of the current module in paths?

crates/mun_hir/src/code_model/module.rs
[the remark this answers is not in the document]
Wodann: How do you ensure that sub-module diagnostics are output though?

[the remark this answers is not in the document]
baszalmstra: The code that deals with that, iterates over all modules in all packages.

crates/mun_hir/src/ty/tests.rs
baszalmstra: I added them.

crates/mun_hir/src/code_model/module.rs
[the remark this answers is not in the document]
baszalmstra: See: https://github.com/mun-lang/mun/pull/283/files#diff-31fbab3d02f7bf5b413c5d15d6f429883c4c185fc8c0cb96509ff4701a260f23R205

crates/mun_syntax/src/tests/parser.rs
Wodann: I'm not super opinionated about it. I always found it interesting that there were two ways to do the same thing in Rust (although one might be more explicit than the other), as it is likely to cause confusion. "If there are two ways, am I doing it the right way?" 🤔 

For example:
```rust
struct Bar;

mod foo {
  pub struct Bar;
}

use foo::Bar;
```

Doing this results in a compile error:
```bash
   Compiling playground v0.0.1 (/playground)
error[E0255]: the name `Bar` is defined multiple times
 --> src/main.rs:7:5
  |
1 | struct Bar;
  | ----------- previous definition of the type `Bar` here
...
7 | use foo::Bar;
  |     ^^^^^^^^ `Bar` reimported here
  |
  = note: `Bar` must be defined only once in the type namespace of this module
help: you can use `as` to change the binding name of the import
  |
7 | use foo::Bar as OtherBar;
  |     ^^^^^^^^^^^^^^^^^^^^
```

so there is no actual (mandatory) use case for using the `self::Bar`. (I'm assuming here that the argument extends to a use-statement of sub-modules)

crates/mun_hir/src/item_scope.rs
```diff
@@ -0,0 +1,87 @@
...
+    types: FxHashMap<Name, (ItemDefinitionId, Visibility)>,
+
+    /// All values visible in this scope
+    values: FxHashMap<Name, (ItemDefinitionId, Visibility)>,
+
+    /// All items that are defined in this scope
+    defs: Vec<ItemDefinitionId>,
+}
+
+pub(crate) static BUILTIN_SCOPE: Lazy<FxHashMap<Name, PerNs<(ItemDefinitionId, Visibility)>>> =
```
Wodann, line 20: As we're renaming, should this be `PRIMITIVE` and `PrimitiveType::ALL`?

crates/mun_hir/src/item_tree/lower.rs
```diff
@@ -1,9 +1,10 @@
 //! This module implements the logic to convert an AST to an `ItemTree`.
 
 use super::{
-    Field, Fields, Function, IdRange, ItemTree, ItemTreeData, ItemTreeNode, LocalItemTreeId,
-    ModItem, Struct, StructDefKind, TypeAlias,
+    diagnostics, Field, Fields, Function, IdRange, ItemTree, ItemTreeData, ItemTreeNode,
+    LocalItemTreeId, ModItem, RawVisibilityId, Struct, StructDefKind, TypeAlias,
 };
+use crate::visibility::RawVisibility;
 use crate::{
```
Wodann, line 8: Can you merge these use-statements?

crates/mun_hir/src/module_tree.rs
```diff
@@ -0,0 +1,227 @@
...
+// Using a `LocalModuleId` you can access the `ModuleTree` to get the `ModuleData`
+impl std::ops::Index<LocalModuleId> for ModuleTree {
+    type Output = ModuleData;
+    fn index(&self, id: LocalModuleId) -> &ModuleData {
+        &self.modules[id]
+    }
+}
+
+impl ModuleTree {
+    /// Construct the tree of modules from the set of files in a package
```
Wodann, line 42:
```suggestion
    /// Constructs the tree of modules from the set of files in a package.
```

crates/mun_hir/src/module_tree.rs
```diff
@@ -0,0 +1,227 @@
...
+            } else {
+                None
+            }
+        })
+    }
+}
+
+/// Given a path name, returns the module name
+fn normalize_module_name(name: Component) -> String {
+    name.as_str().replace("-", "_")
```
Wodann, line 145: Is `-` allowed in paths?

crates/mun_hir/src/module_tree.rs
```diff
@@ -0,0 +1,227 @@
...
+}
+
+/// Given a relative path, returns a Vec with all the module names
+fn path_to_module_path(path: &RelativePath) -> Vec<String> {
+    if path.extension().is_none() {
+        path.components().map(normalize_module_name).collect()
+    } else if path
+        .file_stem()
+        .map(|stem| stem.to_lowercase())
+        .expect("the file has an extension so it must also have a file stem")
```
Wodann, line 155: What happens if I have a file `.mun`? Would this have an empty stem?

Wodann (changes requested):
The next batch of reviewed files 😅

crates/mun_syntax/src/tests/parser.rs
baszalmstra: I get your point. I mainly use it to indicate that something is comming from this or one of the child modules. Mostly in use statements. In `use` statements you always have to start with the package name OR a local module. I use `self` here to be clear that an import name is relative to the current module instead of being a crate. Also if there is a child module with the same name as a package it makes sense to use it (vs `super::<self_name>::foo`)

I do still see some usecases that makes this useful so if you dont mind too much I'll leave it in.

crates/mun_hir/src/item_scope.rs
baszalmstra: I did leave the `BUILTIN_SCOPE` (because thats what it actually is), but I did rename `BuiltinType` to `PrimitiveType`

crates/mun_hir/src/module_tree.rs
baszalmstra: It is on disk if thats whay you mean?

crates/mun_hir/src/module_tree.rs
baszalmstra: From the docs:
```
/// The stem is:
///
/// * [None], if there is no file name;
/// * The entire file name if there is no embedded `.`;
/// * The entire file name if the file name begins with `.` and has no other `.`s within;
/// * Otherwise, the portion of the file name before the final `.`
```

So then the stem would be `.mun`, Ill add an error for invalid module names.

crates/mun_syntax/src/tests/parser.rs
Wodann: I think avoiding conflicts with crate/package names is a very good point.
In that case, is there value to always requiring `self` to avoid this confusion; or just leave it up to developers to decide for themselves?

crates/mun_hir/src/module_tree.rs
Wodann: My point is that if path names directly translate to module names, we should just mandate snake_case?

I think Rust originally allowed `-`, but later changed that. We can choose to just keep it consistent.

crates/mun_syntax/src/tests/parser.rs
baszalmstra: I do see a usecase for that in `use` statements but I wouldnt want to always have to type self there.

crates/mun_hir/src/module_tree.rs
baszalmstra: I changed it so that it no longer normalized the path and just errors on paths with a `-`.

baszalmstra:
I think I solved all issues, also rebased everything on current master.

crates/mun_hir/src/name_resolution/path_resolution.rs
```diff
@@ -0,0 +1,149 @@
...
+        module: LocalModuleId,
+        name: &Name,
+    ) -> PerNs<(ItemDefinitionId, Visibility)> {
+        self[module]
+            .get(name)
+            .or(BUILTIN_SCOPE.get(name).copied().unwrap_or_else(PerNs::none))
+    }
+
+    /// Resolves the specified `path` from within the specified `module`. Also returns whether or
+    /// not an additions to the PackageDef would change the result. This is useful when resolving
```
Wodann, line 73:
```suggestion
    /// not additions to the `PackageDef` would change the result. This is useful when resolving
```

crates/mun_hir/src/name_resolution/path_resolution.rs
```diff
@@ -0,0 +1,149 @@
...
+    /// all imports.
+    pub(super) fn resolve_path_with_fixedpoint(
```
Wodann, line 75: Based on the description, I'm not sure why this function has `fixedpoint` in its name.

crates/mun_hir/src/name_resolution/path_resolution.rs
```diff
@@ -0,0 +1,149 @@
...
+            PathKind::Package => PerNs::types((
+                ModuleId {
+                    package: self.module_tree.package,
+                    local_id: self.module_tree.root,
+                }
+                .into(),
+                Visibility::Public,
+            )),
+            PathKind::Abs => {
+                // TODO: How do we get here?
```
Wodann, line 115: If so, can this be an `unreachable!`?

crates/mun_hir/src/package_defs.rs
```diff
@@ -0,0 +1,32 @@
+mod collector;
+
+use crate::ids::PackageId;
```
Wodann, line 3: If possible, can you merge these use statements?

crates/mun_hir/src/package_defs/collector.rs
```diff
@@ -0,0 +1,200 @@
+use super::PackageDefs;
+use crate::arena::map::ArenaMap;
```
Wodann, line 2: If possible, can you merge these use statements?

crates/mun_hir/src/package_defs/collector.rs
```diff
@@ -0,0 +1,200 @@
...
+        db,
+        package_id,
+        modules: Default::default(),
+        module_tree: db.module_tree(package_id),
+    };
+    collector.collect();
+    collector.finish()
+}
+
+struct DefCollector<'db> {
```
Wodann, line 26: Missing struct-level docs.

crates/mun_hir/src/package_defs/collector.rs
```diff
@@ -0,0 +1,200 @@
...
+            parent: Option<(Name, LocalModuleId)>,
+        ) {
+            // Insert an empty item scope for this module, this will be filled in.
+            collector.modules.insert(module_id, ItemScope::default());
+
+            // If there is a file associated with the module, collect all definitions from it
+            let module_data = &collector.module_tree[module_id];
+            if let Some(file_id) = module_data.file {
+                let item_tree = collector.db.item_tree(file_id);
+                ModCollector {
```
Wodann, line 51: This notation throws me off a little as it seems like you are doing something to a temporary that is then destroyed (and resulting data lost). In reality, probably the `ModCollector` is a wrapper around the `collector` that ensures the data is stored in `collector`.
Could we add a function to the type of `collector` that wraps does this instead?

Wodann (changes requested):
The last batch of comments. Hopefully we can merge this soon 🌲

crates/mun_hir/src/name_resolution/path_resolution.rs
baszalmstra: I modified the comment so that this is a bit clearer.

crates/mun_hir/src/name_resolution/path_resolution.rs
baszalmstra: I removed the `PathKind::Abs` altogether. I don't see a use-case for it.

crates/mun_hir/src/package_defs/collector.rs
baszalmstra: I introduced a variable. I hope that makes it a bit clearer.

Wodann (approved):

crates/mun_hir/src/package_defs/collector.rs
Wodann: I don't know that it helps solves the issue I pointed out. To me, it's still unclear that you are actually operating on the `DefCollector` through the helper struct. If the goal was to use helper functions that operate on `DefCollector`s public API, I'd probably just do so using unassociated `pub fn`s instead of creating a helper struct that obfuscates the control flow.

I might be overlooking something; or it's a personal preference, so I'm not making an issue of this.

baszalmstra: I hear you. I tried a couple of approaches:

- I moved everything to seperate function but then the argument list becomes quite large and it feels like code duplication
- I moved the code you highlighted to a seperate function inside the DefCollector (still keeping the ModCollector) but it requires a lot of parameters that feel very out of context in the new function. I could move the entire if statement to a function but that feels just as out of place to me.
- I tried keeping it the same but only remove the collector from the struct, but then a lot of lifetimes no longer work which doesnt improve the readibility of the code.

Would it help if I renamed the `ModCollector` to `ModCollectionContext`?

Wodann: I think a context would work. You'd then pass that context to (unassocitaed) methods that use it?

baszalmstra: I renamed the `ModCollector` to `ModCollectorContext`, I then made all its functions unassociated but that seemed even more weird to me because there is a lot of duplication. So I kept the `impl` of the `ModCollectorContext`.

Wodann (approved):

crates/mun_hir/src/package_defs/collector.rs
Wodann: Thanks for all of your effort trying to improve this! 🙂

PR #286, opened 2020-10-23
feat(codegen): add AsValue macro support for enums

### LINKED ISSUE #275: Support for enums in the `AsValue` macro
The [`AsValue` macro](https://github.com/mun-lang/mun/blob/master/crates/mun_codegen_macros/src/lib.rs) that is used for procedurally generating LLVM IR conversion functions for Rust types currently only supports `struct`s.

We also want to be able to generate conversion functions for `enum`s, as follows:

```rust
#[derive(AsValue)]
#[repr(u8)]
pub enum TypeInfoData<'ink> {
    Primitive,
    Struct(StructInfo<'ink>),
}
```

Wodann:
Closes #275

### DIFF
--- a/crates/mun_codegen/src/ir/types.rs
+++ b/crates/mun_codegen/src/ir/types.rs
@@ -1,4 +1,6 @@
-use crate::value::{AsValue, IrValueContext, SizedValueType, TransparentValue, Value};
+use crate::value::{
+    AsValue, BytesOrPtr, IrTypeContext, IrValueContext, SizedValueType, TransparentValue, Value,
+};
 use itertools::Itertools;
 use mun_codegen_macros::AsValue;
 
@@ -8,6 +10,10 @@ impl<'ink> TransparentValue<'ink> for abi::Guid {
     fn as_target_value(&self, context: &IrValueContext<'ink, '_, '_>) -> Value<'ink, Self::Target> {
         self.0.as_value(context)
     }
+
+    fn as_bytes_and_ptrs(&self, _: &IrTypeContext<'ink, '_>) -> Vec<BytesOrPtr<'ink>> {
+        vec![self.0.to_vec().into()]
+    }
 }
 
 impl<'ink> TransparentValue<'ink> for abi::Privacy {
@@ -15,6 +21,10 @@ impl<'ink> TransparentValue<'ink> for abi::Privacy {
 
     fn as_target_value(&self, context: &IrValueContext<'ink, '_, '_>) -> Value<'ink, Self::Target> {
         (*self as u8).as_value(context)
+    }
+
+    fn as_bytes_and_ptrs(&self, _: &IrTypeContext<'ink, '_>) -> Vec<BytesOrPtr<'ink>> {
+        vec![vec![*self as u8].into()]
     }
 }
 
@@ -24,6 +34,10 @@ impl<'ink> TransparentValue<'ink> for abi::TypeGroup {
     fn as_target_value(&self, context: &IrValueContext<'ink, '_, '_>) -> Value<'ink, Self::Target> {
         (*self as u8).as_value(context)
     }
+
+    fn as_bytes_and_ptrs(&self, _: &IrTypeContext<'ink, '_>) -> Vec<BytesOrPtr<'ink>> {
+        vec![vec![*self as u8].into()]
+    }
 }
 
 impl<'ink> TransparentValue<'ink> for abi::StructMemoryKind {
@@ -31,6 +45,10 @@ impl<'ink> TransparentValue<'ink> for abi::StructMemoryKind {
 
     fn as_target_value(&self, context: &IrValueContext<'ink, '_, '_>) -> Value<'ink, Self::Target> {
         (self.clone() as u8).as_value(context)
+    }
+
+    fn as_bytes_and_ptrs(&self, _: &IrTypeContext<'ink, '_>) -> Vec<BytesOrPtr<'ink>> {
+        vec![vec![self.clone() as u8].into()]
     }
 }
 
--- a/crates/mun_codegen/src/value/string.rs
+++ b/crates/mun_codegen/src/value/string.rs
@@ -1,4 +1,4 @@
-use super::{AsValue, Global, IrValueContext, TransparentValue, Value};
+use super::{AsValue, BytesOrPtr, Global, IrTypeContext, IrValueContext, TransparentValue, Value};
 use std::ffi::{CStr, CString};
 
 /// Enables internalizing certain data structures like strings.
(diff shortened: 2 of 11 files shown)


crates/mun_codegen/src/ir/types.rs
```diff
@@ -7,6 +8,10 @@ impl<'ink> TransparentValue<'ink> for abi::Guid {
     fn as_target_value(&self, context: &IrValueContext<'ink, '_, '_>) -> Value<'ink, Self::Target> {
         self.0.as_value(context)
     }
+
+    fn as_bytes_and_ptrs(&self, _: &IrValueContext<'ink, '_, '_>) -> Vec<BytesOrPtr<'ink>> {
```
baszalmstra, line 12: I think the function name is a bit weird. It returns a bytes *or* ptr, but the function is called bytes *and* ptr.

crates/mun_codegen/src/ir/types.rs
```diff
@@ -7,6 +8,10 @@ impl<'ink> TransparentValue<'ink> for abi::Guid {
...
+        vec![BytesOrPtr::Bytes(self.0.to_vec())]
```
baszalmstra, line 13: Would it make sense to implement a `From<Vec<u8>>` for `BytesOrPtr`?

baszalmstra (commented):
First little batch, Im gonna wait until this is rebased to get a better overview of the changes. :)

crates/mun_codegen/src/ir/types.rs
Wodann: It returns a vector of `BytesOrPtr`, which means it can contain both. We want pointers to remain pointers; anything else needs to be converted to bytes. That makes it a function that converts to bytes and pointers; hence the name.

crates/mun_codegen/src/ir/types.rs
Wodann: That is actually implemented. I'll change this instance.

Wodann:
I need to test trivial tests for `as_bytes_and_ptrs` as the test coverage is too low.

crates/mun_codegen/src/ir/types.rs
baszalmstra: 👍

crates/mun_codegen/src/value/string.rs
```diff
@@ -57,6 +57,12 @@ impl<'ink> TransparentValue<'ink> for CString {
     fn as_target_value(&self, context: &IrValueContext<'ink, '_, '_>) -> Value<'ink, Self::Target> {
         self.as_bytes_with_nul().as_value(context)
     }
+
+    fn as_bytes_and_ptrs(&self, _: &IrValueContext<'ink, '_, '_>) -> Vec<BytesOrPtr<'ink>> {
+        unreachable!(
+            "`as_bytes_and_ptrs` should never be called on a `String`, as it cannot be a member of an enum."
```
baszalmstra, line 63:
```suggestion
            "`as_bytes_and_ptrs` should never be called on a `String`, as its size is dynamic."
```

crates/mun_codegen/src/value/string.rs
```diff
@@ -65,4 +71,10 @@ impl<'ink> TransparentValue<'ink> for String {
     fn as_target_value(&self, context: &IrValueContext<'ink, '_, '_>) -> Value<'ink, Self::Target> {
         self.as_bytes().as_value(context)
     }
+
+    fn as_bytes_and_ptrs(&self, _: &IrValueContext<'ink, '_, '_>) -> Vec<BytesOrPtr<'ink>> {
+        unreachable!(
+            "`as_bytes_and_ptrs` should never be called on a `String`, as it cannot be a member of an enum."
```
baszalmstra, line 77:
```suggestion
            "`as_bytes_and_ptrs` should never be called on a `String`, as its size is dynamic."
```

crates/mun_codegen_macros/src/lib.rs
```diff
@@ -173,37 +247,82 @@ pub fn as_value_derive(input: TokenStream) -> TokenStream {
                         let type_context = context.type_context;
                         let field_padding = #field_padding_values;
 
-                        let struct_type = Self::get_ir_type(context.type_context);
+                        if <#ident>::has_const_value() {
```
baszalmstra, line 250: What does this mean in this context? It would be good to add a comment here on the two paths.

crates/mun_codegen_macros/src/lib.rs
```diff
@@ -213,8 +332,370 @@ pub fn as_value_derive(input: TokenStream) -> TokenStream {
         Data::Union(_) => {
             unimplemented!("#[derive(AsValue)] is not defined for unions!");
         }
-        Data::Enum(_) => {
-            unimplemented!("#[derive(AsValue)] is not defined for enums!");
+        Data::Enum(enum_data) => {
+            const SUPPORTED_TAG_SIZES: &[&str] =
```
baszalmstra, line 336: Does it make sense to split these two into sperate functions? One for struct and one for enum?

crates/mun_codegen_macros/src/lib.rs
```diff
@@ -213,8 +332,370 @@ pub fn as_value_derive(input: TokenStream) -> TokenStream {
...
+                &["u8", "u16", "u32", "u64", "i8", "i16", "i32", "i64"];
+
+            let repr_ty = derive_input.attrs.iter().find_map(|a| {
```
baszalmstra, line 339: What does this do exactly?

crates/mun_codegen_macros/src/lib.rs
```diff
@@ -213,8 +332,370 @@ pub fn as_value_derive(input: TokenStream) -> TokenStream {
...
+                    #ident
+                }
+            } else {
+                // Default to u32
+                quote! {
+                    u32
+                }
+            };
+
+            if enum_data.variants.is_empty() {
```
baszalmstra, line 383: Should we add a todo on the backlog for this?

baszalmstra (changes requested):
This is really nice and also provides a valuable start to add enums to Mun itself! 

In general I think it would be good to split the generation of enums and structs in seperate functions, maybe add some more helper functions. This procedural macro function is now almost 700 lines long which IMO is way to long for a single function. 

Also, again, some comments explaining what is going on would be nice. Maybe you could even add a little explaination in the docs what the problem is and how it is solved? (Either in the book or as doc comments)

crates/mun_codegen_macros/src/lib.rs
Wodann: `proc_macro` is a wrapper around `proc_macro2`. `proc_macro2` is unstable. `proc_macro` is stable. The results of `quote` etc. are all `proc_macro2`, hence why you need to call `.into()`.

That's all legal because you are not actually importing unstable types. However, once you use it as function parameters though, you have to import those types - which is not possible iirc.

crates/mun_codegen_macros/src/lib.rs
Wodann: Does it add value to support an `enum` without variants? E.g. this is no legal in Rust:

```rust
enum Foo {} // legal

fn main() {
    let foo = Foo {}; // illegal: error[E0574]: expected struct, variant or union type, found enum `Foo`
    println!("Hello, world!");
}
```

baszalmstra: I see thats fine then :)

crates/mun_codegen_macros/src/lib.rs
Wodann: I forgot this, I think. If everything else is fine, I'll include this in the final rebase submission

Wodann: Resolved.

baszalmstra (approved):

PR #290, opened 2020-12-29
feature: use statements language support

### LINKED ISSUE #249: Add language support for `use` statement
Similar to Rust, we want to be able to `use` functions, constants and types from other modules. However, our keywords and implementation differ slightly. For more information, see #248.

Given a folder structure:
- mod.mun
- foo.mun
- bar.mun

file: `mod.mun`
```mun
pub use foo::Foo; // Re-export a child's definition

struct Baz;
```

file: `foo.mun`
```mun
use package::{bar::Bar, Baz};
// Illegal: A statement with the same name was already imported (Even though both use statements are equivalent)
use super::Baz; 

pub(super) struct Foo {
    baz: Baz, // Can use private definitions from any of its ancestors
}

// Illegal: cannot leak a definition with smaller scope
// (i.e. [private] < pub(super) < pub(package) < pub)
pub foo_from_bar(bar: Bar) -> Foo {
    bar.0
}
```

file: `bar.mun`
```mun
use package::Foo;

pub(package) Bar(Foo);
```

baszalmstra:
Adds the use of `use` statements in the code. 

#### ✔️ Features

- [x] Parsing of use statements (e.g. `use foo::bar::{Bar, Baz}`)
- [x] Resolve all use statements in all modules
- [x] Errors for unresolved use statement
- [x] Errors for cyclic use statements
- [x] Wildcard use statements
- [x] Proper error for importing a duplicate name

#### ❓ Missing features, but planned

I will create issues for these once this is merged.

- Errors for illegal import statement visibility 
- Errors for leaking non-visibile types

#### ❌ Out of scope

- Runtime dependencies (so this is actually not actually usable yet)

Closes #249

### DIFF
--- a/crates/mun_hir/src/code_model/module.rs
+++ b/crates/mun_hir/src/code_model/module.rs
@@ -1,7 +1,7 @@
 use super::{Function, Package, Struct, TypeAlias};
 use crate::ids::{ItemDefinitionId, ModuleId};
 use crate::primitive_type::PrimitiveType;
-use crate::{DiagnosticSink, FileId, HirDatabase};
+use crate::{DiagnosticSink, FileId, HirDatabase, Name};
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub struct Module {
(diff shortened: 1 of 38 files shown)


crates/mun_hir/src/visibility.rs
```diff
@@ -57,6 +58,23 @@ pub enum Visibility {
 }
 
 impl Visibility {
+    /// Returns true if an item with this visibility is accessible from the module from the
```
Wodann, line 61:
```suggestion
    /// Returns true if an item with this visibility is accessible from the module of the
```

crates/mun_syntax/src/parsing/grammar/declarations.rs
```diff
@@ -105,3 +109,79 @@ fn opt_fn_ret_type(p: &mut Parser) -> bool {
...
+                _ => (),
+            }
+        }
+        _ => {
+            m.abandon(p);
+            let msg = "expected one of `self`, `super`, `package` or an identifier";
+            if top_level {
+                p.error_recover(msg, DECLARATION_RECOVERY_SET);
+            } else {
+                // if we are parsing a nested tree, we have to eat a token to main balanced `{}`
```
Wodann, line 154:
```suggestion
                // if we are parsing a nested tree, we have to eat a token to remain balanced `{}`
```

Wodann: ?

Wodann (changes requested):
Nicely done. No comments about the architecture or code; just a couple of requests in terms of spelling, grammar, and documentation.

crates/mun_hir/src/code_model/module.rs
```diff
@@ -23,6 +23,49 @@ impl Module {
...
+                local_id: parent_id,
+            },
+        })
+    }
+
+    /// Returns the name of this module or None if this is the root module
+    pub fn name(self, db: &dyn HirDatabase) -> Option<Name> {
+        let module_tree = db.module_tree(self.id.package);
+        let parent = module_tree[self.id.local_id].parent?;
+        module_tree[parent]
```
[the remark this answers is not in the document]
baszalmstra: The parent has a mapping from name to child.

Wodann (approved):

PR #293, opened 2021-01-03
feat: adds lsp document symbol provider

baszalmstra:
Adds an LSP document symbol provider. The result is based on the AST of a SourceFile. The implementation is adapted from the rust-analyzer code.

image

PR #294, opened 2021-01-11
feat(language server): integrated new vfs

baszalmstra:
This quite a large PR that drops the use `ra_vfs` and instead implements a custom solution based on the current solution in rust-analyzer but a tailored towards Mun.

This also adds a new crate `mun_paths` that provides convenience structs for dealing with absolute and relative paths. SourceRoots have also been changed a bit. They now contain the relative paths to the files they contain instead of that being stored in the database. Since a single file might be contained in multiple source roots this makes more sense. I also introduced the `PackageSet` to the database which holds all different packages that the database should know of.

A lot of boilerplate around the language server is now consolidated which now also enables multiple mun packages in a single workspace. 

I also hacked a bit around the use of async functions. Im planning to remove all async code and using a simple taskpool. I think async works really well for IO operations but for our computationally heavy usecase its easier to use threads. (did this in #295)

This PR also paves the way for much better testing of the language server because its now much easier to reason about the virtual filesystem and its progress (which was missing in the previous implementation). (similar to #293 )

### DIFF
--- a/crates/mun_language_server/src/lib.rs
+++ b/crates/mun_language_server/src/lib.rs
@@ -7,14 +7,17 @@ mod conversion;
 mod db;
 mod diagnostics;
 mod main_loop;
+mod project_manifest;
 pub mod protocol;
+mod workspace;
 
 pub use config::Config;
 pub use main_loop::main_loop;
 
-use crate::config::FilesWatcher;
-use serde::de::DeserializeOwned;
-use serde::Serialize;
+use crate::{config::FilesWatcher, project_manifest::ProjectManifest};
+use paths::AbsPathBuf;
+use serde::{de::DeserializeOwned, Serialize};
+use std::convert::TryFrom;
 
 pub type Result<T> = anyhow::Result<T>;
 
@@ -67,8 +70,24 @@ pub async fn run_server_async() -> Result<()> {
     }
 
     let config = {
-        let mut config = Config::default();
+        // Convert the root uri to a PathBuf
+        let root_dir = match initialize_params
+            .root_uri
+            .and_then(|it| it.to_file_path().ok())
+            .and_then(|path| AbsPathBuf::try_from(path).ok())
+        {
+            Some(path) => path,
+            None => {
+                // Get the current working directory as fallback
+                let cwd = std::env::current_dir()?;
+                AbsPathBuf::try_from(cwd)
+                    .expect("could not convert current directory to an absolute path")
+            }
+        };
 
+        let mut config = Config::new(root_dir);
+
+        // Determine type of watcher to use
         let supports_file_watcher_dynamic_registration = initialize_params
             .capabilities
             .workspace
@@ -79,24 +98,26 @@ pub async fn run_server_async() -> Result<()> {
             config.watcher = FilesWatcher::Client;
         }
 
-        // Get the current working directory as fallback
-        let cwd = std::env::current_dir()?;
-        // Convert the root uri to a PathBuf
-        let root = initialize_params
-            .root_uri
-            .and_then(|it| it.to_file_path().ok())
-            .unwrap_or(cwd);
         // Convert the workspace_roots, if these are empy use the root_uri or the cwd
-        config.workspace_roots = initialize_params
+        let workspace_roots = initialize_params
             .workspace_folders
             .map(|workspaces| {
                 workspaces
                     .into_iter()
                     .filter_map(|it| it.uri.to_file_path().ok())
+                    .filter_map(|path| AbsPathBuf::try_from(path).ok())
                     .collect::<Vec<_>>()
             })
             .filter(|workspaces| !workspaces.is_empty())
-            .unwrap_or_else(|| vec![root]);
+            .unwrap_or_else(|| vec![config.root_dir.clone()]);
+
+        // Find all the projects in the workspace
+        let discovered = ProjectManifest::discover_all(workspace_roots.iter().cloned());
+        log::info!("discovered projects: {:?}", discovered);
+        if discovered.is_empty() {
+            log::error!("failed to find any projects in {:?}", workspace_roots);
+        }
+        config.discovered_projects = Some(discovered);
 
         config
     };
--- a/crates/mun_language_server/src/project_manifest.rs
+++ b/crates/mun_language_server/src/project_manifest.rs
@@ -1,0 +1,62 @@
+use anyhow::bail;
+use paths::{AbsPath, AbsPathBuf};
+use rustc_hash::FxHashSet;
+use std::convert::TryFrom;
+use std::fs::read_dir;
+use std::io;
+
+/// A wrapper around a path to a mun project
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Ord, PartialOrd)]
+pub struct ProjectManifest {
+    pub path: AbsPathBuf,
+}
+
+impl ProjectManifest {
+    /// Constructs a new [`ProjectManifest`] from a path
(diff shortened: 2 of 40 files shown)


crates/mun_language_server/src/lib.rs
```diff
@@ -79,24 +98,26 @@ pub async fn run_server_async() -> Result<()> {
...
-            .unwrap_or(cwd);
         // Convert the workspace_roots, if these are empy use the root_uri or the cwd
-        config.workspace_roots = initialize_params
+        let workspace_roots = initialize_params
             .workspace_folders
             .map(|workspaces| {
                 workspaces
                     .into_iter()
                     .filter_map(|it| it.uri.to_file_path().ok())
+                    .filter_map(|path| AbsPathBuf::try_from(path).ok())
```
Wodann, line 108: Can't this be a single `a(path).ok() && b(path).ok()`?

crates/mun_paths/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "mun_paths"
+version = "0.1.0"
+authors = ["Bas Zalmstra <zalmstra.bas@gmail.com>"]
```
Wodann, line 4: This should be the Mun team

crates/mun_paths/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+edition = "2018"
+description = "Provides convenience structures for handling relative- and absolute paths"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
Wodann, line 8:
```suggestion
```

crates/mun_vfs/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "mun_vfs"
+version = "0.1.0"
+authors = ["Bas Zalmstra <zalmstra.bas@gmail.com>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
Wodann, line 7:
```suggestion
```

crates/mun_vfs/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "mun_vfs"
+version = "0.1.0"
+authors = ["Bas Zalmstra <zalmstra.bas@gmail.com>"]
```
Wodann, line 4: Should be the Mun team

Wodann (commented):
Reviewed the first half of this change. In general it looks good, but I'm not sure why we need to move away from `async`? Was that just easier to keep in line with RA?

baszalmstra:
> but I'm not sure why we need to move away from `async`? Was that just easier to keep in line with RA?
Its not about performance but more about the general simplicity of the code. It also makes it easier to use some of the tools out there. A lot is not async (salsa being a big example). You can see currently there are a lot of block_ons in there for that reason.

crates/mun_language_server/src/lib.rs
baszalmstra: No 'ok' returns an Option. We use the result of the first operation in the next

crates/mun_language_server/src/project_manifest.rs
```diff
@@ -0,0 +1,62 @@
+use anyhow::{bail, Result};
```
Wodann, line 1: I'm generally not a fan of shadowing `std` types that are in the `std` prelude, as it's impossible for anyone reading the code to know

crates/mun_language_server/src/project_manifest.rs
```diff
@@ -0,0 +1,62 @@
...
+
+/// A wrapper around a path to a mun project
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Ord, PartialOrd)]
+pub struct ProjectManifest {
+    pub path: AbsPathBuf,
+}
+
+impl ProjectManifest {
+    /// Constructs a new [`ProjectManifest`] from a path
+    pub fn from_manifest_path(path: impl AsRef<AbsPath>) -> Result<Self> {
```
Wodann, line 16:
```suggestion
    pub fn from_manifest_path(path: impl AsRef<AbsPath>) -> anyhow::Result<Self> {
```

crates/mun_vfs/src/lib.rs
```diff
@@ -0,0 +1,235 @@
...
+    pub fn file_path(&self, file_id: FileId) -> &AbsPath {
+        self.interner.lookup(file_id)
+    }
+
+    /// Returns the content of the file with the specified `FileId`.
+    pub fn file_contents(&self, file_id: FileId) -> Option<&[u8]> {
+        self.get(file_id).as_deref()
+    }
+
+    /// Returns an iterator that iterates over all the `FileId`s and their path.
```
Wodann, line 83:
```suggestion
    /// Returns an iterator that iterates all `FileId`s and their path.
```

crates/mun_vfs/src/lib.rs
```diff
@@ -0,0 +1,235 @@
...
+            .enumerate()
+            .filter(|(_, contents)| contents.is_some())
+            .map(move |(id, _)| {
+                let file_id = FileId(id as u32);
+                let path = self.interner.lookup(file_id);
+                (file_id, path)
+            })
+    }
+
+    /// Notify this instance that the contents of the specified file has changed to something else.
```
Wodann, line 96:
```suggestion
    /// Notifies this instance that the contents of the specified file has changed to something else.
```

Wodann (commented):
Still 2 large files left to review

crates/mun_language_server/src/project_manifest.rs
baszalmstra: Agree, fixed.

crates/mun_vfs/src/monitor/notify_monitor.rs
```diff
@@ -0,0 +1,290 @@
...
+        entry: MonitorEntry,
+        watch: bool,
+    ) -> Vec<(AbsPathBuf, Option<Vec<u8>>)> {
+        match entry {
+            MonitorEntry::Files(files) => self.load_files_entry(files, watch),
+            MonitorEntry::Directories(dirs) => self.load_directories_entry(dirs, watch),
+        }
+    }
+
+    /// Loads all the files and optionally add to watched entries
```
Wodann, line 187:
```suggestion
    /// Loads all the files and optionally adds to watched entries
```

Wodann (changes requested):
Once you've incorporated these changes, feel free to merge.

crates/mun_vfs/src/monitor.rs
```diff
@@ -0,0 +1,211 @@
...
+    fn set_config(&mut self, config: MonitorConfig);
+
+    /// Reload the content of the specified file. This will trigger a new `Loaded` message to be
+    /// send.
+    fn reload(&mut self, path: &AbsPath);
+}
+
+impl MonitorDirectories {
+    /// Returns true if, according to this instance, the file at the given `path` is contained in
+    /// this set.
```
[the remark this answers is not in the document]
baszalmstra: I actually think `returns true if` is clearer. Do you want me to replace the commas with dashes?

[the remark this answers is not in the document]
Wodann: The intend was to replace the "true if" as whether is the linguistic form of representing that. I always find it weird when people refer to a logical (or programmer) perspective of `true` in sentences, such as in the documentation.

Wodann (approved):

PR #296, opened 2021-01-12
feat(cli): add option to emit IR

Wodann:
Adds the `--emit-ir` CLI flag to the `build` subcommand, for outputting LLVM IR to `*.ll` files instead of generating Mun libraries.

This PR also removes the need for `#[serial]` tests by adding and using a `path` input argument for the `init` and `new` subcommands, for deterministic execution of tests - in parallel.


baszalmstra (approved):
Nice and clean! 👍

PR #295, opened 2021-01-11
feat(language server): removed async code and switched to lsp_server 

baszalmstra:
Depends on #294 

This PR removes all async code and switch to the use of the `lsp_server` crate instead of using our own.

PR #297, opened 2021-01-14
feature(lsp): adds document symbol provider 

baszalmstra:
Adds an LSP document symbol provider. The result is based on the AST of a SourceFile. The implementation is adapted from the rust-analyzer code.

This is a take 2 it depends on #295 . Its a much cleaner approach using the previous PRs. It supersedes #293 .

image


crates/mun_language_server/src/handlers.rs
```diff
@@ -0,0 +1,48 @@
...
+            range: convert_range(symbol.node_range, &line_index),
+            selection_range: convert_range(symbol.navigation_range, &line_index),
+            children: None,
+        };
+        parents.push((doc_symbol, symbol.parent));
+    }
+
+    // Builds hierarchy from a flat list, in reverse order (so that indices
+    // makes sense)
+    let document_symbols = {
```
Wodann, line 31: It seems like this might have its own function & tests?

crates/mun_language_server/src/state.rs
```diff
@@ -375,6 +377,23 @@ impl LanguageServerSnapshot {
 
         Ok(url)
     }
+
+    /// Converts the specified `Url` to a `hir::FileId`
```
Wodann, line 381: I see url and uri used interchangeably. Why's that?

crates/mun_language_server/src/state/protocol/dispatcher.rs
```diff
@@ -41,6 +41,36 @@ impl<'a> RequestDispatcher<'a> {
         Ok(self)
     }
 
+    /// Try to dispatch the event as the given Request type on the thread pool.
+    pub fn on<R>(
```
Wodann, line 45:
```suggestion
    pub fn on_request<R>(
```

Wodann: It seems like the subject of "on" is missing. Is the subject the request?

crates/mun_language_server/src/state/protocol/dispatcher.rs
```diff
@@ -41,6 +41,36 @@ impl<'a> RequestDispatcher<'a> {
...
+        &mut self,
+        f: fn(LanguageServerSnapshot, R::Params) -> anyhow::Result<R::Result>,
```
Wodann, line 47: What kind of function is `f`? Maybe use that as a parameter name?

Wodann (changes requested):

crates/mun_language_server/src/state.rs
baszalmstra: Its because LSP does the same. A text document is referred to by (copied from lsp_types):
```
/// The text document's URI.
pub uri: Url,
```

crates/mun_language_server/src/state/protocol/dispatcher.rs
baszalmstra: The subject is `R`. It's written like: `.on::<lsp_types::request::DocumentSymbolRequest>(handlers::handle_document_symbol)?`

crates/mun_language_server/src/state/protocol/dispatcher.rs
baszalmstra: Renamed to `compute_response_fn`, also in different places.

crates/mun_language_server/src/state/protocol/dispatcher.rs
Wodann: Either works

Wodann (approved):
Up to you to decide whether you think it's necessary to rename `on` to `on_request`

PR #305, opened 2021-01-28
feat: add mut keyword

### LINKED ISSUE #199: [Tracking] Variable mutability
Similar to Rust, we want to use opt-in mutability; i.e. by default all variables are immutable. If you want to edit a variable, you need to declare it as with the `mut` keyword. E.g. this example should result in a compile error:

```rust
let foo = 5;
if condition() {
  foo += 1; // error: `foo` is not mutable
}
```
This code would be correct:
```rust
let mut foo = 5;
if condition() {
  foo += 1;
}
```

mkhan45:
Adds the `mut` keyword and updates a lot of tests to accomodate for it. 

Closes #199.

:tada:

### DIFF
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -24,3 +24,6 @@
 - [Developer Documentation](ch04-00-developer-docs.md)
     - [Salsa](ch04-01-salsa.md)
     - [Building LLVM](ch04-02-building-llvm.md)
+
+- [RFCs](rfcs/rfcs.md)
+    - [Dynamically sized arrays](rfcs/0001-dynamically-sized-arrays.md)
--- a/book/src/rfcs/0001-dynamically-sized-arrays.md
+++ b/book/src/rfcs/0001-dynamically-sized-arrays.md
@@ -1,0 +1,165 @@
+- Feature name: dynamically_sized_arrays
+- Start date: 2021-04-23
+- RFC PR: [https://github.com/mun-lang/mun/pull/324](https://github.com/mun-lang/mun/pull/324)
+- RFC Issue: [https://github.com/mun-lang/mun/issues/328](https://github.com/mun-lang/mun/issues/328)
+
+# Summary
+
+This is an RFC to introduce the concept of dynamically sized arrays in Mun.
+These are arrays where the length of the array is not yet known at compile time.
+This is different from statically sized arrays, where the size of the array is known at compile time.
+
+# Motivation
+
+Reasons for having dynamically allocated arrays.
+
+* Having only statically sized arrays limits the use of Mun to only use types which have a known size. 
+* Dynamically sized arrays pose more flexibility than statically sized arrays. 
+    See C# and Swift as an example where statically sized arrays are rarely used.
+* Tuples can already be used for statically sized arrays (although not very ergonomically).
+
+    ```rust,ignore
+    struct(value) ArrayOfFiveFloats(f32,f32,f32,f32,f32)
+    ```
+* Dynamically sized arrays are easily understandable from a user perspective
+
+# Detailed design
+
+This RFC proposes to add the language construct of a dynamically sized array as well as several additions to the language which are required as supporting features.
+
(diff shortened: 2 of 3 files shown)


crates/mun_codegen/src/ir/body.rs
```diff
@@ -102,7 +102,7 @@ impl<'db, 'ink, 't> BodyIrGenerator<'db, 'ink, 't> {
             let body = self.body.clone(); // Avoid borrow issues
 
             match &body[*pat] {
-                Pat::Bind { name } => {
+                Pat::Bind { name, mode: _ } => {
```
baszalmstra, line 105: I think using `..` is a bit nicer here.
```suggestion
                Pat::Bind { name, .. } => {
```

crates/mun_codegen/src/ir/body.rs
```diff
@@ -528,7 +528,7 @@ impl<'db, 'ink, 't> BodyIrGenerator<'db, 'ink, 't> {
         };
 
         match &self.body[pat] {
-            Pat::Bind { name } => {
+            Pat::Bind { name, mode: _ } => {
```
baszalmstra, line 531: Same here
```suggestion
            Pat::Bind { name, .. } => {
```

crates/mun_hir/src/diagnostics.rs
```diff
@@ -273,6 +273,32 @@ impl Diagnostic for InvalidLHS {
     }
 }
 
+#[derive(Debug)]
+pub struct ImmutableLHS {
```
baszalmstra, line 277: Maybe something like `ReassignmentToImmutableBinding` or something?

crates/mun_hir/src/ty/infer.rs
```diff
@@ -320,6 +321,12 @@ impl<'a> InferenceResultBuilder<'a> {
                                 lhs: *lhs,
                             })
                         }
+                        if !self.check_mut_bind(&resolver, *lhs) {
```
baszalmstra, line 324: I think it makes sense to turn this into an `else if` so that when you have something like `3=5` you don't get two errors.

crates/mun_hir/src/ty/infer/mut_bind.rs
```diff
@@ -0,0 +1,30 @@
...
+use std::sync::Arc;
+
+impl<'a> InferenceResultBuilder<'a> {
+    /// Checks if the specified expression is a place-expression. A place expression represents a
+    /// memory location.
+    pub(super) fn check_mut_bind(&mut self, resolver: &Resolver, expr: ExprId) -> bool {
+        let body = Arc::clone(&self.body); // avoid borrow checker problem
+        match &body[expr] {
+            Expr::Path(p) => self.check_mut_path(resolver, p),
+            Expr::Field { expr, name: _ } => self.check_mut_bind(resolver, *expr),
```
baszalmstra, line 14:
```suggestion
            Expr::Field { expr, .. } => self.check_mut_bind(resolver, *expr),
```

crates/mun_syntax/src/parsing/grammar/patterns.rs
```diff
@@ -37,6 +34,9 @@ fn placeholder_pat(p: &mut Parser) -> CompletedMarker {
 
 fn bind_pat(p: &mut Parser) -> CompletedMarker {
     let m = p.start();
+    if p.at(T![mut]) {
+        p.bump(T![mut]);
+    }
```
baszalmstra, line 39:
```suggestion
    p.eat(T![mut]);
```

crates/mun_syntax/src/tests/parser.rs
```diff
@@ -120,6 +120,29 @@ fn expression_statement() {
...
+    }
+    "#,
+    )
+}
+
+#[test]
+fn mutable_argument() {
+    snapshot_test(
+        r#"
+    fn foo(mut a) {
```
baszalmstra, line 139: Missing type after `a`

baszalmstra (changes requested):
Thanks for this great PR! I've added some cosmetic comments but other than that it looks good to me!

baszalmstra (approved):

mkhan45:
I didn't mean to close this; I think I updated the wrong branch on my side. I'll fix it tomorrow.

PR #324, opened 2021-04-23
[RFC] dynamically sized arrays

baszalmstra:
This is an RFC for dynamically sized arrays in Mun.

[Rendered](https://deploy-preview-324--docs-mun-lang.netlify.app/rfcs/0001-dynamically-sized-arrays.html)

### DIFF
--- a/book/src/rfcs/0001-dynamically-sized-arrays.md
+++ b/book/src/rfcs/0001-dynamically-sized-arrays.md
@@ -1,6 +1,7 @@
 - Feature name: dynamically_sized_arrays
 - Start date: 2021-04-23
 - RFC PR: [https://github.com/mun-lang/mun/pull/324](https://github.com/mun-lang/mun/pull/324)
+- RFC Issue: [https://github.com/mun-lang/mun/issues/328](https://github.com/mun-lang/mun/issues/328)
 
 # Summary
 
@@ -86,7 +87,7 @@ When you add elements to an array and that array begins to exceed its reserved c
 > let array = [0.0; count] 
 >
 > // constructs an array of `count` elements all initialized to `Foo {}`
-> let array = [value; count]
+> let array = [Foo{}; count]
 >
 > // constructs an array of `count` elements all initialized to `value`
 > let array = [value; count]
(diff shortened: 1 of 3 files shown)


AngelicosPhosphoros:
I don't use Mun in production yet, I just a stranger.

Would you like to support something like Python generator syntax?
```python
let my_array = [
   x*x for x in range(100)
]
```

book/src/rfcs/0001-dynamically-sized-arrays.md
```diff
@@ -0,0 +1,162 @@
...
+> let array = [i32]::with_capacity(some_initial_size);
+> ```
+
+> TODO: In the future it would be nice if you can create an array by replicating a certain expression.
+> ```rust,ignore
+> // constructs an array of `count` elements all initialized to 0.0
+> let array = [0.0; count] 
+>
+> // constructs an array of `count` elements all initialized to `Foo {}`
+> let array = [value; count]
```
Wodann, line 89:
```suggestion
> let array = [Foo; count]
```

Wodann:
I like the proposal!

As we're looking to be C ABI compatible, we'll likely need to add support for static arrays at some point (e.g. if we want to generate Mun bindings for a C header). The current proposal leaves room for a more specific `[f32;8]` to specify fixed-size arrays :)

baszalmstra:
> I don't use Mun in production yet, I just a stranger.
> 
> Would you like to support something like Python generator syntax?
> 
> ```python
> let my_array = [
>    x*x for x in range(100)
> ]
> ```

That would be very nice! But maybe at a later stage. I first want to get “primitive” support in there.

baszalmstra:
> I like the proposal!
> 
> As we're looking to be C ABI compatible, we'll likely need to add support for static arrays at some point (e.g. if we want to generate Mun bindings for a C header). The current proposal leaves room for a more specific `[f32;8]` to specify fixed-size arrays :)
> 
> By default type elision can still use `[f32]`. If you want a fixed-size array, you just need to add the type. If we want to support both fixed- and dynamically sized arrays, we'd likely need a "slice" alternative though.
> 
> One option we've discusses is C#'s https://docs.microsoft.com/en-us/dotnet/api/system.arraysegment-1
> 
> If we want a shorthand we could use:
> 
> * `&[f32]` for any size of slice
> * `&[f32;8]` for a slice of size 8

Yes I intentionally left that open as an option. Ill add that in the document. And indeed something like an ArraySegment would be very useful then.

Wodann (approved):

Issue #309, opened 2021-02-26
Can currently leak private types and values


baszalmstra:
Reexporting a type or value from one module to another should check the visibility of the original item.

This can happen in a few cases:

```rust
//- foo.mun
// Illegal, Bar has a smaller scope that this use statement
pub(super) struct Bar;

// Illegal, Bar has a smaller scope than this function
pub fn baz() -> Bar {
}
```

```rust
//- mod.mun
pub use foo::Bar; // Should be illegal because Bar is only visible to this module
```

In rust this emits the error: `can't leak private type`. This is currently not checked in Mun.

mkhan45:
We also need to check that a public struct does not contain any private types in its exported fields:

```rs
struct Foo;

pub struct Bar {
    pub foo: Foo // illegal since Foo has a smaller scope than Bar
}
```

mkhan45:
this is closed by #338 and #344

baszalmstra:
Nice! Thanks @mkhan45 !

PR #338, opened 2021-07-15
Function privacy leak validator

mkhan45:
Part of #309 

Adds a validation for leaking private types from function signatures

### DIFF
--- a/crates/mun_diagnostics/src/hir/exported_private.rs
+++ b/crates/mun_diagnostics/src/hir/exported_private.rs
@@ -1,0 +1,48 @@
+use super::HirDiagnostic;
+use crate::{Diagnostic, SourceAnnotation};
+use mun_syntax::{AstNode, TextRange};
+
+/// An error that is emitted when trying to leak a private type
+pub struct ExportedPrivate<'db, 'diag, DB: mun_hir::HirDatabase> {
+    _db: &'db DB,
+    diag: &'diag mun_hir::diagnostics::ExportedPrivate,
+    value_name: String,
+}
+
+impl<'db, 'diag, DB: mun_hir::HirDatabase> Diagnostic for ExportedPrivate<'db, 'diag, DB> {
+    fn range(&self) -> TextRange {
+        self.diag.highlight_range()
+    }
+
+    fn title(&self) -> String {
+        format!("can't leak `{}`", self.value_name)
+    }
+
+    fn primary_annotation(&self) -> Option<SourceAnnotation> {
+        Some(SourceAnnotation {
+            range: self.diag.highlight_range(),
+            message: self.diag.message(),
+        })
+    }
+}
+
+impl<'db, 'diag, DB: mun_hir::HirDatabase> ExportedPrivate<'db, 'diag, DB> {
+    /// Constructs a new instance of `ExportedPrivate`
+    pub fn new(db: &'db DB, diag: &'diag mun_hir::diagnostics::ExportedPrivate) -> Self {
+        let parse = db.parse(diag.file);
+
+        // Get the text of the value as a string
+        let value_name = diag
+            .type_ref
+            .to_node(&parse.syntax_node())
+            .syntax()
+            .text()
+            .to_string();
+
+        ExportedPrivate {
+            _db: db,
+            diag,
+            value_name,
+        }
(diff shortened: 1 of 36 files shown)


crates/mun_hir/src/expr/validator.rs
```diff
@@ -40,6 +42,40 @@ impl<'a> ExprValidator<'a> {
...
+            Visibility::Module(module_id) => {
+                ty.visibility(self.db).is_visible_from(self.db, module_id)
+            }
+            Visibility::Public => ty.visibility(self.db).is_externally_visible(),
+        };
+        let type_is_not_allowed = |ty: &Ty| !type_is_allowed(ty);
+
+        param_types
+            .chain(std::iter::once(ret_ty))
+            .filter(type_is_not_allowed)
```
mkhan45, line 69: add .take(1) to show only a single diagnostic per function

baszalmstra:
Thanks for this!

Would it be possible to highlight the region of the type thas has a more private type? So in the case of:

```
pub fn foo(ty: PrivateType) { 
               ~~~~~~~~~~~ <-- this part is highlighted: cant leak `PrivateType` 
 /// ...
}
```

mkhan45:
Alright, I've changed the diagnostic to highlight the type, but it's a bit harder to make it actually display the type name in the diagnostic message. I don't think it's a big deal since the name is highlighted anyway though.

```
error: can't leak private type
 --> mod.mun:3:16
  |
3 | pub fn foo(ty: PrivateType) -> PrivateType {
  |                ^^^^^^^^^^^ can't leak private type
  |error: can't leak private type
 --> mod.mun:3:32
  |
3 | pub fn foo(ty: PrivateType) -> PrivateType {
  |                                ^^^^^^^^^^^ can't leak private type
```

baszalmstra:
You can implement something in `mun_diagnostic` to also be able to give proper feedback on the name. In that crate you conversion happens from `mun_hir::Diagnostic`s to `mun_diagnostic::Diagnostic`. A lot more context is known at that point (like the entire HIR database) which should make it fairly easy to get the name of the type. See https://github.com/mun-lang/mun/blob/master/crates/mun_diagnostics/src/hir/unresolved_type.rs as an example.

Adding the name is not required perse but I think it will provide users with a much better experience down the road.

baszalmstra:
Also works properly with the language server: 

image

Nice!

mkhan45:
I've added a `mun_diagnostic` diagnostic so the type names show: https://github.com/mun-lang/mun/pull/338/files#diff-f6d9d08feaa1fdc33c5467abad598c17045b35ab9dc93d19b1bf0acf8e61fa6c

baszalmstra (approved):
Nice! 👍

crates/mun_hir/src/diagnostics.rs
```diff
@@ -178,6 +178,26 @@ impl Diagnostic for ExpectedFunction {
...
 
+#[derive(Debug)]
+pub struct ExportedPrivate {
+    pub file: FileId,
+    pub type_ref: AstPtr<ast::TypeRef>,
+}
+
+impl Diagnostic for ExportedPrivate {
+    fn message(&self) -> String {
+        "can't leak private type".to_string()
```
Wodann, line 189: We have duplication of this error message. Is there a way to reuse it to avoid code+paste errors?

Other occurance:
https://github.com/mun-lang/mun/pull/338/files#diff-7a57632f7acc9a0b1abad5ef6f9447f95fabecf03312d0ab7d0597807ae8e25dR24

crates/mun_diagnostics/src/hir/exported_private.rs
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    fn title(&self) -> String {
+        format!("can't leak `{}`", self.value_name)
+    }
+
+    fn primary_annotation(&self) -> Option<SourceAnnotation> {
+        Some(SourceAnnotation {
+            range: self.diag.highlight_range(),
+            message: self.diag.message(),
```
mkhan45, line 24: Reuses the error message of the interior struct

mkhan45:
I've updated it so that there's no duplication

Wodann (approved):

PR #341, opened 2021-07-25
feat: early support for arrays

baszalmstra:
Adds support for arrays.

Missing feature:
- [x] Hotreloading of arrays and types in arrays: https://github.com/baszalmstra/mun/pull/20
- [x] Using arrays from the runtime
- [x] Documentation in the form of a book entry

Out of scope 
- Adding and removing entries
- Querying the length


baszalmstra:
Depends on #268

baszalmstra:
> I would really like to support creating arrays from a Rust iterator. Something like:
> 
> ```rust
> let mun_array: ArrayRef = runtime.construct_array::<i32>(vec![1,2,3,4,].into_iter())
> ```
> 
> However, this requires type references, because the type information of the array cannot be created. The type information has to be created and shared with the mun libraries. This requires type references.

It took almost a year, but this is now possible in this branch!

crates/mun_memory/src/gc/mark_sweep.rs
```diff
@@ -115,19 +116,243 @@ where
...
+            self.next = unsafe { NonNull::new_unchecked(self.next.as_ptr().add(self.stride)) };
+            Some(element)
+        } else {
+            None
+        }
+    }
+}
+
+/// Creates a layout describing the record for `n` instances of `layout`, with a suitable amount of
+/// padding between each to ensure that each instance is given its requested size an alignment.
```
Wodann, line 283:
```suggestion
/// padding between each to ensure that each instance is given its requested size and alignment.
```

crates/mun_memory/src/gc/mark_sweep.rs
```diff
@@ -551,15 +798,20 @@ enum Color {
...
+    pub data: ObjectInfoData,
     pub roots: u32,
     pub color: Color,
     pub ty: Arc<TypeInfo>,
 }
 
+#[repr(C)]
+union ObjectInfoData {
+    pub ptr: *mut u8,
+    pub array: *mut ArrayHeader,
```
Wodann, line 812: Do all objects store the pointer now?

crates/mun_memory/src/type_info.rs
```diff
@@ -130,6 +148,22 @@ impl TypeInfo {
...
             mutable_pointer_type: Default::default(),
+            array_type: Default::default(),
+        })
+    }
+
+    /// Constructs a new array type
+    pub fn new_array(element: Arc<TypeInfo>) -> Arc<TypeInfo> {
+        Arc::new(TypeInfo {
+            name: format!("[{}]", &element.name),
+            layout: Layout::new::<*const std::ffi::c_void>(),
```
Wodann, line 159: Is there a reason you use `*const c_void`?

Wodann (changes requested):
WIP review

crates/mun_codegen/src/ir.rs
```diff
@@ -1,22 +1,29 @@
...
 pub mod file;
 pub(crate) mod file_group;
 pub mod function;
 mod intrinsics;
+mod reference;
 pub mod ty;
 pub(crate) mod type_table;
 pub mod types;
 
+use array::MunArrayValue;
```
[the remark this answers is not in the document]
baszalmstra: I wanted to indicate that this is how an Array is represented in the Mun runtime. Would RuntimeArrayValue perhaps be a better name?

crates/mun_codegen/src/ir/array.rs
```diff
@@ -0,0 +1,121 @@
...
+//! }
+//! ```
+
+use crate::ir::reference::MunReferenceValue;
+use inkwell::builder::Builder;
+use inkwell::types::{BasicTypeEnum, IntType, StructType};
+use inkwell::values::{BasicValueEnum, IntValue, PointerValue};
+use std::ffi::CStr;
+
+/// A helper struct that wraps a PointerValue which points to an in memory Mun array value.
```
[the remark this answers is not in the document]
baszalmstra: A `MunReferenceValue` represents a value that is stored in the GC heap. The helper structs enables easier (and safer) access to the data stored via the reference. Both an array and a GC struct are represented using such a reference.

crates/mun_codegen/src/ir/array.rs
```diff
@@ -0,0 +1,121 @@
...
+        self.0.get_name()
+    }
+
+    /// Generate code to get to the array value.
+    fn get_array_ptr(&self, builder: &Builder<'ink>) -> PointerValue<'ink> {
+        self.0.get_data_ptr(builder)
+    }
+
+    /// Generate code to fetch the length of the array.
+    pub fn get_length_ptr(&self, builder: &Builder<'ink>) -> PointerValue<'ink> {
```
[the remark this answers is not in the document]
baszalmstra: The length can be modified at runtime, the capacity cannot. Therefor I need mutable access to the length (hence the pointer) but I don't for the capacity. Probably adding a `get_length` fn will make sense when we need it.

crates/mun_memory/src/gc/mark_sweep.rs
baszalmstra: No this is a `union`, so depending on the type stored in the ObjectInfo either the `ptr` or the `array` is valid.

crates/mun_memory/src/type_info.rs
baszalmstra: Actually, it doesn't really make sense to have a layout for an array. The layout has a double meaning:

There are two types of layouts:
1. The layout to allocate the object on the heap.
2. The layout of a reference to a value of type T.

Layout in this context means the first, we store how many bytes (and with which alignment) the data for a type needs on the GC heap. For GC structs that is different than (2) because to store a reference another layout is needed (a pointer layout).

For arrays (1) doesn't make any sense because it depends both on the layout of the elements as well as the capacity of the array. We should probably refactor this at some point to make a clear distinction between the two.

crates/mun_memory/src/gc/mark_sweep.rs
```diff
@@ -1,62 +1,179 @@
...
+                return Some(event);
             }
         }
         None
     }
 }
 
+/// A struct that enables iterating over all GC references in a struct.
+///
+/// TODO: if the element type doesnt contain any references its a bit of a waste to iterate over all
```
Wodann, line 147:
```suggestion
/// TODO: if the element type doesnt contain any references it's a bit of a waste to iterate over all
```

crates/mun_memory/src/gc/mark_sweep.rs
```diff
@@ -109,9 +226,213 @@ where
...
+
+/// Helper object to work with GcPtr that represents an array.
+///
+/// Arrays are stored in memory with a header which holds the length and capacity. The memory layout
+/// of an array looks like this in memory:
+///
+/// ```text
+/// object.data.array ───►┌──────────────┐
+///                       │ ArrayHeader  │
+///                       └─┬────────────┘
```
Wodann, line 263: Start of a book entry?

crates/mun_memory/src/gc/mark_sweep.rs
```diff
@@ -109,9 +226,213 @@ where
...
+            next,
+            stride: element_layout.pad_to_align().size(),
+        }
+    }
+}
+
+/// An iterator implementation.
+///
+/// TODO: Note that this iterator is highly non-thread safe. Any operation that modifies the
+/// original array could case undefined behavior.
```
Wodann, line 372:
```suggestion
/// original array could cause undefined behavior.
```

crates/mun_runtime/src/lib.rs
```diff
@@ -445,6 +475,54 @@ impl Runtime {
...
+    where
+        I::IntoIter: ExactSizeIterator,
+    {
+        let iter = iter.into_iter();
+        let element_type = T::type_info();
+        let array_type = element_type.array_type();
+        let array_capacity = iter
+            .size_hint()
+            .1
+            .expect("iterator doesnt return upper bound");
```
Wodann, line 493:
```suggestion
            .expect("iterator doesn't return upper bound");
```

crates/mun_runtime/tests/arrays.rs
```diff
@@ -0,0 +1,116 @@
...
+    pub fn main() -> [i32] { [5,4,3,2,1] }
+    ",
+        |builder| builder,
+    )
+    .expect("Failed to build test driver");
+
+    let result: ArrayRef<'_, i32> = driver.runtime.invoke("main", ()).unwrap();
+
+    assert_eq!(result.len(), 5);
+    assert!(result.capacity() >= 5);
```
Wodann, line 18: Why not equal to 5?

Wodann (changes requested):
Last comments :)

crates/mun_memory/src/type_info.rs
Wodann: I agree. For arrays we could just use a `const` layout, saving memory. Same applies for pointers (in the future I expect). Could you outline this in a new issue?

crates/mun_codegen/src/ir/array.rs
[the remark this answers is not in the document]
Wodann: Thanks. Is that reflected in its documentation?

crates/mun_codegen/src/ir.rs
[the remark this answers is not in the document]
Wodann: I think so. My interpretation would be that it's a wrapper around the inkwell ArrayValue, as opposed to a runtime specific one.

crates/mun_memory/src/type_info.rs
baszalmstra: See #443

crates/mun_memory/src/gc.rs
```diff
@@ -1,9 +1,11 @@
+mod array;
 mod mark_sweep;
 mod ptr;
 mod root_ptr;
 
 use crate::r#type::Type;
 use std::marker::PhantomData;
+use std::ptr::NonNull;
```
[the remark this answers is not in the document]
baszalmstra: I've been thinking about this and I would prefer to leave these types of formatting as is for now until these things become standardized through rustfmt. I don't wait to stray away from default rustfmt because it introduces extra manual work for all contributors that can probably be automated in the future anyway.

crates/mun_runtime/tests/arrays.rs
baszalmstra: Because in the future we might allocate a little more memory to accommodate extra pushing. The requirement on the capacity is that it can at least store `5` elements or more.

crates/mun_codegen/src/ir/array.rs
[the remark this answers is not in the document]
baszalmstra: Yeah there is a big section in the doc comment of `MunReferenceValue`

Wodann:
For the rocket and spaceship logic, can all of tr updates not just happen in Mun? In theory, we only need to query their positions from Rust to draw them

PR #344, opened 2021-08-04
Feat: Added private type leak validation to structs

mkhan45:
Also exposed struct field visibility in the HIR and created a struct validation module

### DIFF
--- a/crates/mun_hir/src/code_model.rs
+++ b/crates/mun_hir/src/code_model.rs
@@ -2,7 +2,7 @@ mod function;
 mod module;
 mod package;
 pub(crate) mod src;
-mod r#struct;
+pub(crate) mod r#struct;
 mod type_alias;
 
 use crate::{expr::BodySourceMap, HirDatabase, Name};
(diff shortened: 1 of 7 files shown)


crates/mun_hir/src/code_model/struct.rs
```diff
@@ -19,6 +19,8 @@ use crate::visibility::RawVisibility;
 pub use ast::StructMemoryKind;
 use std::iter::once;
 
+pub(crate) mod validator;
```
Wodann, line 22: Does this need to be `pub(crate)`?

crates/mun_hir/src/code_model.rs
```diff
@@ -2,7 +2,7 @@ mod function;
 mod module;
 mod package;
 pub(crate) mod src;
-mod r#struct;
+pub(crate) mod r#struct;
```
Wodann, line 5: Does this need to be `pub(crate)`?

Wodann (approved):

PR #349, opened 2021-09-13
fixed bench compilation, but they don't run

mkhan45:
The benchmarks stopped compiling after #347. This commit fixes bench compilation but right now it still panics on each invoke call with a 'failed to obtain function' error

### DIFF
--- a/crates/mun_runtime/benches/benchmarks.rs
+++ b/crates/mun_runtime/benches/benchmarks.rs
@@ -1,5 +1,5 @@
 use criterion::{black_box, criterion_group, criterion_main, BenchmarkId, Criterion};
-use mun_runtime::{invoke_fn, StructRef};
+use mun_runtime::StructRef;
 
 mod util;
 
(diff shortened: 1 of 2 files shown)


Wodann:
To prevent this from happening in the future, could you please specify to test benchmarks: `cargo t --all-targets` in the CI config?

source: https://doc.rust-lang.org/stable/cargo/commands/cargo-test.html#target-selection

mkhan45:
Should I add it to the commit hook as well?

Wodann:
> Should I add it to the commit hook as well?

That'd be great, yes. Thank you!

Wodann:
> Should I add it to the commit hook as well?

That'd be great, yes. Thank you!

Wodann:
> Should I add it to the commit hook as well?

That'd be great, yes. Thank you!

baszalmstra:
I fixed all issues in #359

Issue #198, opened 2020-05-17
Logo?


baszalmstra:
We currently don't really have a logo other than a cutout of the text on the website. It would be really awesome if we could design a proper logo for Mun that we could use all over the place.

evolbug:
Hey, made something for fun, since the language takes inspiration around Lua and Rust
![mun](https://user-images.githubusercontent.com/19688801/82198636-42a1af00-9905-11ea-90a4-56575ecd7481.png)

Wodann:
@evolbug thanks for making the logo! I really like your direction and the incorporation of the moon and Rust cog.

I have one request. Could you try an inverse design, with the moon forming the perimeter and the cog design (somehow) incorporated on the inside? I hope that that makes the design a little less "busy"

evolbug:
@Wodann I don't think it'd look any less busy, it's also harder to incorporate the rust gear and moon in the inverse order. It would need either element changed entirely then

erlend-sh:
Some possible variations (obviously just showing reference pictures here, not original work):

Just the moon, without the cogs, but with the darker pattern of the moon not-so-subtly in the shape of a rustacean.
![image](https://user-images.githubusercontent.com/583842/82366449-7b7b7a00-9a12-11ea-8c92-b017885784a4.jpeg)

A rustacean in front of the moon
![](https://image.freepik.com/free-vector/deer-moon-logo_74472-68.jpg)

Or on top

![](https://veectori.com/wp-content/uploads/edd/2020/02/s4-2048x2048.jpg)

Wodann:
Thanks for the mood board. I really like the ideas 🙂

Wodann:
> @Wodann I don't think it'd look any less busy, it's also harder to incorporate the rust gear and moon in the inverse order. It would need either element changed entirely then

Sorry, I missed your response. Yes, my idea would be to leave the moon as is, but incorporate a smaller version of the gear (or multiple) on the surface of the moon. We don't per se have to stick with an identical gear to Rust - not at all actually. A reference would be good enough, similar to how we have a reference to Lua by using the moon.

baszalmstra:
Mun actually comes from the Moon (or Mun) in [kerbal space program](https://wiki.kerbalspaceprogram.com/wiki/Mun). 😁

We also have to watch out that the logo (or a version of it) also works as a small icon that we can use in editors etc.

Wodann:
I tried editing @evolbug's logo to illustrate what I meant:

[...]

PR #355, opened 2021-11-23
docs(readme): add logo

jonatcln:
Adds the potential new logo of Mun to the README. This also makes it easier to discuss #198 (i.e. any adjustment that need to be made).

Points that require further thought:

- Is the logo/shape OK?
- What color should the logo be? What is Mun's branding color?
- Should there be a logotype variant of the logo (one that includes the name Mun)?
- Where should the reference logo be stored?


Wodann:
Thank you so much for putting this together.

Those are great questions! @baszalmstra and I are having a call to work on Mun on December 11th. If you don't mind, we'll discuss the questions then and follow up afterwards 🙂

baszalmstra:
I also added this logo to the website, here is the PR: https://github.com/mun-lang/website/pull/84

Wodann:
> Is the logo/shape OK?

Yes, it looks perfect!

> What color should the logo be? What is Mun's branding color?

The current colour scheme works! Would you also be able to make the inverse (i.e. replace black with white, and vice versa) to support dark mode. Github's Markdown actually [supports specifying images for dark/light mode](https://docs.github.com/en/github/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#specifying-the-theme-an-image-is-shown-to).

> Should there be a logotype variant of the logo (one that includes the name Mun)?

If you would be able to make one with the name "Mun" offset to the right, we would really appreciate that! This is the font that we use: https://public-sans.digital.gov/

For reference, you can have a look at this [PR](https://deploy-preview-84--mun-lang.netlify.app/), but feel free to design the layout to your liking 🙂 

> Where should the reference logo be stored?

The `assets` folder that you've created works for us :)

jonatcln:
> Github's Markdown actually supports specifying images for dark/light mode.

Cool! Didn't know that. I added a white version for both the logo and the logotype. GitHub doesn't seem to allow these light/dark-mode-only variants in html img tags, so I couldn't specify the image height in the README but instead embedded it in the svg itself. I thought it would therefore be best to separate the actual logo(type)s (without specific height) from the ones in the README. In order to distinguish them I organized the logo(type)s in two subfolders: `assets/logo` and `assets/readme`.

Let me know if you have any other suggestions :slightly_smiling_face:

jonatcln:
I also quickly made this 'blueprint' of the design:

logotype-blueprint

I'm not sure what to do with it, but it might be handy for future reference.

Also, fun fact: most of the dimensions (the circle-based shapes in particular) originate from this variant of the Fibonacci sequence: 48, 48, 72, 120, 192, 312, 504, ... (because the ratio of two consecutive Fibonacci numbers approximates the golden ratio)

baszalmstra:
It looks incredible! Thank you so much 😊

baszalmstra (approved):

Wodann (approved):

baszalmstra:
Thanks so much! :)

PR #356, opened 2021-12-08
chore(lld): Declare new darwin linker

ovr:
Hello!

LLVM's lld has two different flavors for linking. In this PR I've introduced support for a new one because the old one doesn't work with arm64.

Thanks

### DIFF
--- a/crates/mun_codegen/src/linker.rs
+++ b/crates/mun_codegen/src/linker.rs
@@ -134,7 +134,7 @@ impl Linker for Ld64Linker {
     }
 
     fn finalize(&mut self) -> Result<(), LinkerError> {
-        mun_lld::link(mun_lld::LldFlavor::MachO, &self.args)
+        mun_lld::link(mun_lld::LldFlavor::DarwinOld, &self.args)
             .ok()
             .map_err(LinkerError::LinkError)
     }
--- a/crates/mun_lld/build.rs
+++ b/crates/mun_lld/build.rs
@@ -347,6 +347,9 @@ fn main() {
         println!("cargo:rustc-link-lib=dylib={}", name);
     }
 
+    // xar is used in lld::macho::BitcodeBundleSection
+    println!("cargo:rustc-link-lib=dylib=xar");
+
     let use_debug_msvcrt = env::var_os(format!(
         "LLVM_SYS_{}_USE_DEBUG_MSVCRT",
         env!("CARGO_PKG_VERSION_MAJOR")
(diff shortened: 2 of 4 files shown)


Wodann:
Thanks for the submission. In what version of LLD was this added? We might have to be careful when we merge this as we're still resolving issues with our LLVM upgrade

ovr:
>Thanks for the submission. In what version of LLD was this added? 

As I saw from the source code, starting from 11. In the upcoming 14 version, the old linker for Darwin will be deprecated/removed.

https://github.com/llvm/llvm-project/blob/release/11.x/lld/tools/lld/lld.cpp#L154

crates/mun_lld/build.rs
```diff
@@ -347,6 +347,9 @@ fn main() {
         println!("cargo:rustc-link-lib=dylib={}", name);
     }
 
+    // xar is used in lld::macho::BitcodeBundleSection
+    println!("cargo:rustc-link-lib=dylib=xar");
```
ovr, line 351: *Problem*:

xar is an optional dep, but i didn't find a way how to check on the fly that lld will be compiled with xar.

baszalmstra: mm thats whats causing the build failure currently.. Is the version of lld you are using expecting xar?

crates/mun_codegen/src/linker.rs
```diff
@@ -134,7 +134,7 @@ impl Linker for Ld64Linker {
     }
 
     fn finalize(&mut self) -> Result<(), LinkerError> {
-        mun_lld::link(mun_lld::LldFlavor::MachO, &self.args)
+        mun_lld::link(mun_lld::LldFlavor::DarwinOld, &self.args)
```
baszalmstra, line 137: Maybe rename this to `LegacyDarwin`?

crates/mun_lld/wrapper/lld-c.cpp
```diff
@@ -19,16 +19,18 @@ const char* mun_alloc_str(const std::string& str)
...
 
 extern "C" {
 
 enum LldFlavor {
   Elf = 0,
   Wasm = 1,
-  MachO = 2,
-  Coff = 3,
+  Darwin = 2,
+  DarwinOld = 3,
```
baszalmstra, line 32: Might be nice to add a little comment here explaining what the differences are?

crates/mun_lld/wrapper/lld-c.cpp
```diff
@@ -65,9 +67,15 @@ LldInvokeResult mun_lld_link(LldFlavor flavor, int argc, const char *const *argv
       result.success = lld::wasm::link(args, false, outputStream, errorStream);
       break;
     }
-    case MachO:
+    case Darwin:
     {
-      std::unique_lock <std::mutex> lock(_machOMutex);
+      std::unique_lock <std::mutex> lock(_darwinMutex);
+      result.success = lld::macho::link(args, false, outputStream, errorStream);
```
baszalmstra, line 73: Is the lock here still required or is the new linker reentrant?

baszalmstra:
Our lld implementation has moved to another repository you can find it [here](https://github.com/mun-lang/lld-rs). Im closing this PR here. If you would like to continue, please open a PR there  ( I hope you will 😄  ).

PR #366, opened 2022-02-09
bump: insta 1.12.0

baszalmstra:
With the release of `insta 1.12.0` [a backward-incompatible change was introduced](https://insta.rs/changelog/#1-12-0). This caused a large number of snapshot names to become invalid and undeterministic. This was an obscure feature that we admittedly abused.

I removed the use of `AutoName` from our source code and even moved some of the test results inline, which is a new feature of `insta`. This should reduce the context switch required when viewing snapshots and their corresponding tests since they are now all in one place. I didn't modify any of the algorithms so I assume the output is still valid, I didn't explicitly check this.

This should also fix the current CI failures.


Wodann (commented):
A big reason for not having the expected results inline is that they are hard to create and update if the format changes. Instead `cargo insta` allows you to manage all output files easily.

How has that concern been handled?

baszalmstra:
> A big reason for not having the expected results inline is that they are hard to create and update if the format changes. Instead `cargo insta` allows you to manage all output files easily.
> 
> How has that concern been handled?

Yes "inline snapshots" have been a feature for some time now in insta. Running "cargo insta review" will also update the snapshot in the source code! See https://insta.rs/docs/snapshot-types/#inline-snapshots

Wodann (approved):

PR #370, opened 2022-02-27
feat: make the runtime send

### LINKED ISSUE #299: Using Mun in a threaded environment
The `mun_runtime::Runtime` is not `Sync` nor `Send` since the `RuntimeBuilder::spawn` function returns a `Rc<RefCell<Runtime>>`. This makes it hard to use in other programs that make use of threads. The reason we wrap the `Runtime` in an `Rc` is because the runtime (or better yet, the memory model) is not thread safe. Wrapping the `Runtime` in a mutex will also not work because the memory model is shared between other structs (like `StructRef`). Sending the `Runtime` to a thread and doing modifications while also accessing a `StructRef` from another thread will cause threading issues. 

We've already talked about this a lot. We currently have no proper plan to support multithreading. However, our current solution is also not very ergonomic. I propose we implement a blocking solution that still doesnt do multithreading of the runtime but does allow the runtime and derived data to be shared across threads. In a later version we can then add proper multithreading hopefully using the same API we design now.

A straighforward solution is to wrap all memory modifying operations inside a mutex. This will block any operation performed by other threads but will enable the use of the Runtime and derived data to be send to different threads.

Curious to your thoughts!

baszalmstra:
Removes the need to put the `Runtime` into a `Rc<RefCell<Runtime>>`. This makes the `Runtime` `Send`.

Closes  #299

### DIFF
--- a/crates/mun_runtime/tests/threading.rs
+++ b/crates/mun_runtime/tests/threading.rs
@@ -1,0 +1,5 @@
+use mun_runtime::Runtime;
+
+// Ensures the [`Runtime`] is Send
+trait IsSend: Send {}
+impl IsSend for Runtime {}
--- a/book/listings/ch01-getting-started/listing04.rs
+++ b/book/listings/ch01-getting-started/listing04.rs
@@ -1,20 +1,17 @@
-use mun_runtime::RuntimeBuilder;
+use mun_runtime::Runtime;
 use std::{cell::RefCell, env, rc::Rc};
 
 fn main() {
     let lib_path = env::args().nth(1).expect("Expected path to a Mun library.");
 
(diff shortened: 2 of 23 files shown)


crates/mun_runtime/tests/threading.rs
```diff
@@ -0,0 +1,5 @@
+use mun_runtime::Runtime;
+
+// Ensures the [`Runtime`] is Send
+trait IsSend: Send {}
```
Wodann, line 4: Can't you directly implement `Send` for the `Runtime` class?

Also, should this be in the _tests_ folder?

Wodann (approved):

PR #377, opened 2022-05-15
refactor: link type information at runtime

### LINKED ISSUE #145: Export types from assemblies and store them in the runtime
We want to leave the ownership of the type information with the Garbage Collector so that memory can safely reference its type.

Assemblies will export the type information of types that are defined in them. Everywhere else types are referred to by their GUID. Which will be resolved in the runtime.

assembly type info

### LINKED ISSUE #168: Replace `TypeInfo` with `TypeRef` in Runtime FFI
Currently, the host language is responsible for maintaining the `TypeInfo`'s memory when `insert_fn` is called, but using a `TypeRef` would allow us to avoid this.

Depends on #145

Wodann:
Closes #268 
Closes #145 
Closes #168


cpp/include/mun/struct_ref.h
```diff
@@ -79,7 +81,7 @@ class StructRef {
      *
      * \return a pointer to the struct's type information
      */
-    const MunUnsafeTypeInfo info() const noexcept { return m_runtime->ptr_type(raw()); }
+    const TypeInfo info() const noexcept { return m_runtime->ptr_type(raw()); }
```
baszalmstra, line 84: The `const` here doesnt mean anything.

```suggestion
    TypeInfo info() const noexcept { return m_runtime->ptr_type(raw()); }
```

cpp/include/mun/type_info.h
```diff
@@ -7,71 +7,143 @@
...
-    return &TypeInfo<T>::Type;
-}
+class TypeInfoSpan {
+public:
+    /**
+     * @brief Constructs a type information span from an instantiated `MunTypeInfoSpan`.
+     */
+    TypeInfoSpan(MunTypeInfoSpan span) noexcept : m_span(span) {}
+
+    ~TypeInfoSpan() noexcept { mun_type_info_span_destroy(m_span); }
```
baszalmstra, line 116: Same here, the copy constructor should be deleted.

crates/mun_abi/src/function_info.rs
```diff
@@ -58,67 +57,67 @@ impl FunctionPrototype {
...
-            write!(f, "{}", arg)?;
-        }
-        write!(f, ")")?;
-        if let Some(ret_type) = self.signature.return_type() {
-            write!(f, ":{}", ret_type)?
-        }
-        Ok(())
-    }
-}
+// TODO: Create a runtime-specific version of FunctionPrototype that resolves the `TypeId`s
```
baszalmstra, line 60: Should we remove this commented code?

crates/mun_abi/src/function_info.rs
```diff
@@ -58,67 +57,67 @@ impl FunctionPrototype {
...
+        if self.return_type == <()>::type_info().id {
+            None
+        } else {
+            Some(self.return_type.clone())
         }
-        Ok(())
     }
 }
 
+// impl fmt::Display for FunctionSignature {
```
baszalmstra, line 101: Same, should we remove this commented code?

crates/mun_abi/src/lib.rs
```diff
@@ -48,6 +54,54 @@ pub const SET_ALLOCATOR_HANDLE_FN_NAME: &str = "set_allocator_handle";
...
 pub struct Guid(pub [u8; 16]);
 
+impl From<&[u8]> for Guid {
+    fn from(bytes: &[u8]) -> Self {
+        Guid(md5::compute(&bytes).0)
+    }
+}
+
+#[inline]
+const fn format_hyphenated(src: &[u8; 16]) -> [u8; 36] {
```
baszalmstra, line 64: Does it make sense to add some doc comments here? Maybe move it in the scope of the Display fmt of Guid?

baszalmstra (changes requested):
I left some minor comments, overall this looks good enough. Lets get this merged!

PR #405, opened 2022-05-29
perf: minor optimisations

f01dab1e:


baszalmstra (commented):
Thanks! This is much appreciated. Always good to see some quality of life changes!

One comment: can you change your commit messages to follow: https://www.conventionalcommits.org/en/v1.0.0/  . “minor” is not a valid prefix, I think style or perf would be good here.

Wodann (approved):

PR #415, opened 2022-06-15
Make TypeId as a key of TypeInfo

baszalmstra:
This is an attempt to simplify the runtime and ABI by removing the use of `TypeId` in several places and only using `Arc<TypeInfo>`. This should enable adding pointer type TypeIds.

The idea is that some types are "concrete" which means they are concrete types like a struct or primitive. Pointers (and later arrays) are not concrete types but instead refer indirectly to concrete types.

The tests for the ABI and memory crate succeed, next up is modifying the compiler.


crates/mun_abi/src/lib.rs
```diff
@@ -4,6 +4,21 @@
 //! Runtime.
 #![warn(missing_docs)]
 
+extern crate core;
```
Wodann, line 7: `extern crate` should no longer be necessary since Rust 2018

crates/mun_abi/src/struct_info.rs
```diff
@@ -88,22 +92,81 @@ impl From<StructMemoryKind> for u64 {
     }
 }
 
+impl<'a> PartialEq for StructInfo<'a> {
+    fn eq(&self, other: &Self) -> bool {
```
Wodann, line 96: An interesting question here is: do we want to validate that all fields are equal, or do we want a performant equality operator? Just checking `GUID`s will guarantee that two structs are not the same; and is much faster!

crates/mun_abi/src/type_info.rs
```diff
@@ -40,22 +40,74 @@ impl Debug for TypeInfo {
...
+        s.serialize_field("name", self.name())?;
+        s.serialize_field("size_in_bits", &self.size_in_bits)?;
+        s.serialize_field("alignment", &self.alignment)?;
+        s.serialize_field("data", &self.data)?;
+        s.end()
+    }
+}
+
 /// Contains data specific to a group of types that illicit the same characteristics.
+/// TODO: Can we remove Primitive? It is never exposed through the ABI since its build in.
```
Wodann, line 61: Feel free to, on both accounts

crates/mun_abi/src/type_info.rs
```diff
@@ -229,40 +320,36 @@ macro_rules! impl_primitive_type_info {
...
-        $ty:ty => $name:tt
-    ),+) => {
-        $(
-            impl crate::type_info::HasStaticTypeInfoName for $ty {
-                fn type_name() -> &'static std::ffi::CStr {
-                    static TYPE_INFO_NAME: once_cell::sync::OnceCell<std::ffi::CString> = once_cell::sync::OnceCell::new();
-                    let type_info_name: &'static std::ffi::CString = TYPE_INFO_NAME
-                        .get_or_init(|| std::ffi::CString::new($name).unwrap());
-                    type_info_name.as_ref()
+            impl HasStaticTypeId for $ty {
```
Wodann, line 336: It might be good to create a `type_id.rs` file that contains the struct definition and static type ID-related code

crates/mun_abi/src/type_lut.rs
```diff
@@ -1,36 +1,81 @@
...
-/// Represents a unique identifier for types. The runtime can use this to lookup the corresponding [`TypeInfo`].
-#[repr(C)]
+use itertools::izip;
+
+use crate::Guid;
+
+/// Represents a unique identifier for types. The runtime can use this to lookup the corresponding
+/// [`TypeInfo`]. A [`TypeId`] is a key for a [`TypeInfo`].
+///
+/// A [`TypeId`] only contains enough information to query the runtime for a concrete type.
```
Wodann, line 13: > for a concrete type.

Are we mixing up terminology here? Can't the runtime contain a `*const *const i32`? This is not a concrete type, right?

crates/mun_abi/src/type_lut.rs
```diff
@@ -1,36 +1,81 @@
...
+    Pointer(PointerTypeId<'a>),
+}
+
+/// Represents a pointer to another type.
+#[repr(C)]
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize))]
+pub struct PointerTypeId<'a> {
+    /// The type to which this pointer points
+    pub pointee: &'a TypeId<'a>,
```
Wodann, line 31: Should we create an issue to replace all occurrences of `*const` and `*mut` in `mun_abi` with their reference counterparts?

Wodann (changes requested):
Partial review

crates/mun_abi/src/builtin.rs
```diff
@@ -0,0 +1,122 @@
+///! A module that defines information for built-in (or primitive) types.
+use crate::{Guid, HasStaticTypeId, TypeId};
+
+/// Defines functions for built-in types like f32, i32, etc.
+pub trait BuiltinType: HasStaticTypeId {
```
Wodann, line 5: Previously we referred to these as `Primitive` types. Are we transitioning to built-in?

crates/mun_runtime_capi/src/function_info.rs
```diff
@@ -164,12 +164,12 @@ pub unsafe extern "C" fn mun_function_info_return_type(
 #[cfg(test)]
 pub(crate) mod tests {
     use super::*;
+    use crate::type_info::mun_type_info_eq;
     use crate::{
```
Wodann, line 168: Merge imports

crates/mun_runtime_capi/src/field_info.rs
```diff
@@ -126,13 +126,14 @@ pub unsafe extern "C" fn mun_field_info_span_destroy(span: FieldInfoSpan) -> boo
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::type_info::mun_type_info_eq;
```
Wodann, line 129: Merge imports

crates/mun_runtime/src/reflection.rs
```diff
@@ -129,18 +105,18 @@ where
...
-{
-    fn type_id() -> abi::TypeId {
-        <Self as abi::HasStaticTypeInfo>::type_info().id.clone()
-    }
-
-    fn type_name() -> &'static str {
-        <Self as abi::HasStaticTypeInfo>::type_info().name()
-    }
-}
+// impl<T> ReturnTypeReflection for *const T
```
Wodann, line 108: Commented. Can it be removed?

crates/mun_runtime/src/assembly.rs
```diff
@@ -356,6 +404,12 @@ impl Assembly {
         &self.info
     }
 
+    /// Returns the assembly's information.
+    pub fn info_mut(&mut self) -> &mut abi::AssemblyInfo {
+        // HACK: We want to make sure that the assembly info never outlives self.
```
Wodann, line 409: Why is this necessary now?

crates/mun_runtime/Cargo.toml
```diff
@@ -25,6 +25,8 @@ once_cell = "1.4.0"
 parking_lot = "0.12.0"
 rustc-hash = "1.1"
 seq-macro = "0.3.0"
+ouroboros = "0.15.0"
```
Wodann, line 28: Alphabetic ordering

Wodann (changes requested):
Still WIP

crates/mun_codegen/src/type_info.rs
```diff
@@ -1,244 +1,119 @@
...
-    pub fn from_ir_type<'ink>(ty: &impl AnyType<'ink>, target: &TargetData) -> Self {
-        Self {
-            bit_size: target.get_bit_size(ty),
-            store_size: target.get_store_size(ty),
-            alloc_size: target.get_abi_size(ty),
-            alignment: target.get_abi_alignment(ty),
-        }
-    }
-}
+use abi::static_type_map::StaticTypeMap;
```
Wodann, line 4: Merge imports

crates/mun_codegen/src/ir/type_table.rs
```diff
@@ -1,22 +1,26 @@
...
-use std::{collections::HashMap, convert::TryInto, mem, sync::Arc};
 
 /// A type table in IR is a list of pointers to unique type information that are used to generate
 /// function and struct information.
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub struct TypeTable<'ink> {
-    entries: Vec<TypeInfo>,
-    type_info_to_index: HashMap<TypeInfo, usize>,
+    entries: Vec<Arc<TypeId>>,
+    type_info_to_index: HashMap<Arc<TypeId>, usize>,
```
Wodann, line 23: `type_id_to_index`

crates/mun_codegen/src/ir/type_table.rs
```diff
@@ -103,12 +107,12 @@ impl<'db, 'ink, 't> TypeTableBuilder<'db, 'ink, 't> {
     pub(crate) fn new<'f>(
         db: &'db dyn HirDatabase,
         value_context: &'t IrValueContext<'ink, '_, '_>,
-        intrinsics: impl Iterator<Item = &'f FunctionPrototype>,
+        _intrinsics: impl Iterator<Item = &'f FunctionPrototype>,
```
Wodann, line 110: Why do we no longer need these types?

crates/mun_codegen/src/ir/ty.rs
```diff
@@ -1,8 +1,6 @@
-use crate::{
-    ir::IsIrType,
-    type_info::{TypeInfo, TypeSize},
-};
-use hir::{FloatBitness, HirDatabase, HirDisplay, IntBitness, ResolveBitness, Ty, TyKind};
+use std::sync::Arc;
```
Wodann, line 1: Merge imports

crates/mun_codegen/src/ir/ty.rs
```diff
@@ -11,14 +9,22 @@ use inkwell::{
...
 };
 use smallvec::SmallVec;
-use std::{cell::RefCell, collections::HashMap};
+
+use abi::Guid;
+use hir::{
+    FloatBitness, HirDatabase, HirDisplay, IntBitness, ResolveBitness, Signedness, Ty, TyKind,
+};
+
+use crate::ir::IsIrType;
```
Wodann, line 18: Merge imports

crates/mun_codegen/src/ir.rs
```diff
@@ -192,13 +190,6 @@ impl<'ink, S: BasicType<'ink>, T: IsIrType<'ink, Type = S>> IsPointerType<'ink>
...
 }
 
-// HACK: Manually add `*const TypeInfo`
-impl<'ink> IsPointerType<'ink> for *const TypeInfo {
-    fn ir_type(context: &'ink Context, _target: &TargetData) -> PointerType<'ink> {
-        context.i8_type().ptr_type(AddressSpace::Generic)
-    }
-}
-
 // HACK: Manually add `*const c_void`
```
Wodann, line 193: Why do we still need this one?

crates/mun_codegen/src/code_gen/symbols/ir_type_builder.rs
```diff
@@ -0,0 +1,63 @@
+use std::cell::RefCell;
```
Wodann, line 1: Merge imports

cpp/include/mun/reflection.h
```diff
@@ -11,9 +11,15 @@
...
 namespace mun {
 constexpr inline bool operator==(const MunTypeId& lhs, const MunTypeId& rhs) noexcept {
-    for (auto idx = 0; idx < 16; ++idx) {
-        if (lhs.guid._0[idx] != rhs.guid._0[idx]) {
-            return false;
+    if (lhs.tag != rhs.tag) {
+        return false;
+    }
+
+    if (lhs.tag == MunTypeId_Tag::Concrete) {
```
Wodann, line 18: What if the TypeId is not `Concrete`?

Wodann (changes requested):
First full pass on review is done

crates/mun_abi/src/builtin.rs
baszalmstra: Good one! Both work for me, but for consistency Ill rename this one.

crates/mun_runtime/src/assembly.rs
baszalmstra: Lifetimes are a bit mixed up here. When loading the symbol from the library I would like to already give the AssemblyInfo the lifetime of the library it came from, but I couldn’t get that to compile because it would create a self referential lifetime, since the AssemblyInfo is stored next to the library it came from. To work around this issue I give the stored value the static lifetime and then use this function to ensure the rest of the could does use an AssemblyInfo with a properly scoped lifetime.

crates/mun_codegen/src/ir/type_table.rs
baszalmstra: I think there is still a bug here, really the only types that need to be stored in the type looking tables are the types that we pass to the new function. All the other types can be referred to by typeids.

I’ll update the builder to only insert the types turkey needed.

Wodann: True, given that they are all primitive types.

crates/mun_abi/src/struct_info.rs
baszalmstra: But they don't (yet) include the offsets right?

crates/mun_abi/src/type_lut.rs
baszalmstra: Yes, you were going to create an issue for that right?

crates/mun_abi/src/function_info.rs
```diff
@@ -22,47 +24,47 @@ pub struct FunctionDefinition {
...
     /// Number of argument types
     pub num_arg_types: u16,
 }
 
-/// Owned storage for C-style `FunctionDefinition`.
-pub struct FunctionDefinitionStorage {
-    _name: CString,
-    _arg_types: Vec<TypeId>,
-}
+// /// Owned storage for C-style `FunctionDefinition`.
```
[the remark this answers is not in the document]
baszalmstra: Removed

crates/mun_codegen/src/ir.rs
baszalmstra: We still need these because we cant add a type for `std::ffi::c_void` because it has an undefined type. But the places where we use it, we always use it as a pointer, therefor we only add the information for a pointer to `std::ffi:c_void`.

cpp/include/mun/reflection.h
baszalmstra: I still need to update the C++ code

baszalmstra:
I updated the c++ code as well! This all seems to pass CI now too!

Can you check the comments I responded to @Wodann ?

crates/mun_codegen/src/ir.rs
Wodann: In that case, is this a hack or just part of supporting our required intrinsics?

Or is it a hack because we don't deduce the intrinsics types? 🤔

crates/mun_runtime/src/assembly.rs
Wodann: I saw you added a dependency for self-reference. Can that be removed then?

cpp/include/mun/reflection.h
```diff
@@ -10,74 +10,43 @@
...
-
-}  // namespace mun
-
-#include "struct_ref.h"
-
-namespace mun {
-namespace reflection {
-template <typename T>
-std::optional<std::pair<const char*, const char*>> equals_return_type(
+inline std::optional<std::pair<std::string, std::string>> equals_return_type(
```
Wodann, line 40: Would string_view not be cheaper?

cpp/include/mun/reflection.h
```diff
@@ -10,74 +10,43 @@
...
 
 namespace reflection {
-template <typename T, typename U>
-constexpr bool equal_types() noexcept {
-    return ReturnTypeReflection<T>::type_id() == ReturnTypeReflection<U>::type_id();
-}
-
 template <typename Arg>
-inline std::optional<std::pair<const char*, const char*>> equals_argument_type(
+inline std::optional<std::pair<std::string, std::string>> equals_argument_type(
```
Wodann, line 29: string_view? Or can arg_type_info not be a reference?

cpp/include/mun/struct_ref.h
```diff
@@ -150,8 +158,15 @@ struct Marshal<StructRef> {
     static void move_to(type value, type* ptr, const TypeInfo& type_info) noexcept {
         // Safety: `type_info_as_struct` is guaranteed to return a value for
```
Wodann, line 159: Move the safety comment down to where `struct_` is used

cpp/include/mun/struct_ref.h
```diff
@@ -163,10 +178,16 @@ struct Marshal<StructRef> {
                              const TypeInfo& type_info) noexcept {
         // Safety: `type_info_as_struct` is guaranteed to return a value for
```
Wodann, line 179: Move safety comment down

cpp/include/mun/struct_ref.h
```diff
@@ -208,15 +231,18 @@ std::optional<T> StructRef::get(std::string_view field_name) const noexcept {
 
     // Safety: `type_info_as_struct` is guaranteed to return a value for
```
Wodann, line 232: Idem

cpp/include/mun/struct_ref.h
```diff
@@ -236,15 +262,18 @@ std::optional<T> StructRef::replace(std::string_view field_name, T value) noexce
...
     // `StructRef`s.
```
Wodann, line 264: Idem

Wodann (changes requested):
Some minor nitpicks. Otherwise looks good to me! Awesome work 😃

crates/mun_runtime/src/assembly.rs
baszalmstra: Yes! Good one!

crates/mun_codegen/src/ir.rs
baszalmstra: Yeah, you're right, Ill remove the comment

cpp/include/mun/reflection.h
baszalmstra: Yes, but that also doesn't specify anything about the lifetime of the contents. This is a generic function with no guarantee about the lifetime of the string data. Therefore I changed this to use a string.

baszalmstra:
I fixed some of your final comments, it looks good to me too! @Wodann can you please review my last comments and code? If you're in agreement, feel free to merge! :)

crates/mun_abi/src/struct_info.rs
baszalmstra: 👍 Ill change it to GUID compare then.

PR #416, opened 2022-06-19
misc(example): added Bevy-Mun example

virgilhuxley:
Note: This example can not be run from `cargo run --example bevy_simple`.  
User must be in `examples/bevy_simple` directory to run.


examples/bevy_simple/src/main.rs
```diff
@@ -0,0 +1,50 @@
...
+        println!("Printing value from `mun_func`: {}", result);
+    }
+}
+
+fn reload_munlib_from_timer(
+    time: Res<Time>,
+    mut timer: ResMut<MunReloadTimer>,
+    mut mun: NonSendMut<MunRuntime>,
+) {
+    if timer.0.tick(time.delta()).just_finished() {
```
baszalmstra, line 47: Im wondering, why did you add a timer here? The `update` function should be lightweight enough to call this on every tick. It internally already does optimizations that ensure this.

examples/bevy_simple/src/main.rs
```diff
@@ -0,0 +1,50 @@
...
+        .add_plugins(MinimalPlugins)
+        .add_plugin(MunPlugin)
+        .run();
+}
+
+struct MunPlugin;
+
+impl Plugin for MunPlugin {
+    fn build(&self, app: &mut App) {
+        app.add_startup_system(setup.exclusive_system())
```
baszalmstra, line 17: It would be nice to explain in the example why the `exclusive_system` is required.

Actually, in general it would be nice to have some comment in here that explains that the runtime is non-sync and that bevy has support for this through `exclusive_system`, `insert_non_send_resource`, and `NonSendMut`.

baszalmstra (commented):
Thanks very much for this! This will be very useful for a lot of people. 

I added a few small comments.

examples/bevy_simple/src/main.rs
virgilhuxley: - I included it because it is what I am using and it is proving to work well for me.
- There is not a performance or mun-related reason for choosing a timer-based update system.
- I can add a comment to the effect.

baszalmstra: Ah I see. The reason I asked is because I think that because this is an example people will likely copy it as a starting point. Since the timer serves no real purpose in the example this is code that just complicates the example. Therefore I would for the sake of showing people how to embed Mun in Bevy, remove it.

examples/bevy_simple/src/main.rs
virgilhuxley: Added additional comments.

examples/bevy_simple/src/main.rs
virgilhuxley: Acknowledged. I will remove it.

examples/bevy_simple/src/main.rs
baszalmstra: Perfect!

baszalmstra (approved):
This is perfect! Thanks so much!

baszalmstra:
@Wodann do you know how to fix the CI issue?

Wodann:
> @Wodann do you know how to fix the CI issue?

Yes, the `examples` directory is intended for **mun** packages only. The corresponding **rust** code should be in `crates/mun_runtime`.

The C++ tests look through the **mun** examples to ensure that all of them are compiled. Since the bevy example has the rust code included at the top-level, it fails to find (and compile) the mun library.

baszalmstra:
I am a bit concerned that the rust examples will be hard to find if we add it to the examples directory of the runtime. When I browse repositories the first (and only really) place I look is the examples folder in the root. Would it be an idea to make subdirectories in the examples folder for mun and rust instead? Or perhaps a symlink?

Wodann:
> I am a bit concerned that the rust examples will be hard to find if we add it to the examples directory of the runtime. When I browse repositories the first (and only really) place I look is the examples folder in the root. Would it be an idea to make subdirectories in the examples folder for mun and rust instead? Or perhaps a symlink?

The normal folder structure of a package is like [this](https://doc.rust-lang.org/stable/cargo/guide/project-layout.html). Tests, benchmarks, and examples live alongside the `src` of a crate. This allows you to directly use the crate - without specifying a dependency.

As we have a mono-repo, we have multiple types of examples: Rust examples for `mun_runtime` but potentially also other crates, C++ examples for our wrapper, and Mun examples. My reasoning was to keep the C++ examples in the `cpp` folder, the Mun examples at the top-level, and the Rust examples in their respective crates - based on the project layout linked earlier.

baszalmstra:
Yeah that also sounds good. I just want to make sure that they are easily findable. Having them 3 folders deep might not be the easiest to find.

.github/workflows/ci.yml
```diff
@@ -60,6 +65,12 @@ jobs:
       - name: Install LLVM
         uses: ./.github/actions/install-llvm
 
+      - name: Install packages
```
Wodann, line 68: If we already run `cargo check`, do we also need to run `cargo test` for all examples?

baszalmstra: Not unless we also have tests in there, which seems weird.

Wodann: I realised that we should actually:
- compile Mun examples
- run Rust examples that marshal the generated `*.munlib`s to validate that the APIs match

This can happen in a separate PR though.

Wodann: The Windows worker runs out of memory when we use `--all-targets` for `cargo test`.

I removed benchmarks. We might want to re-add this as a separate benchmark step in our release CI

Wodann:
@baszalmstra I was finally able to fix the CI. Windows no longer uses a cache and has incremental builds disabled. That allows all tests to be compiled without exceeding disk space.

Could you please have a look and approve if you're happy. If you'd like I can also create a separate PR for the required changes and rebase this PR.

baszalmstra (approved):

PR #421, opened 2022-07-07
feature: adds targets for ios

### LINKED ISSUE #417: Support for iOS
I would like to build a `munlib` for iOS. I have tinkered with the code in `mun_target` and added an ios target, and call mun with `--target x86_64-apple-ios`. Still, the generated lib file is for macOS. Any help with the parts that I need to change?

baszalmstra:
This PR adds experimental targets to run Mun on iOS. It adds the targets:

* `x86_64-apple-ios` 
* `aarch64-apple-ios`
* `aarch64-apple-ios-sim`

These are not officially supported since we don't test them in CI.

I also had to do some modifications in the linker to make sure iOS and iOS-sim are properly targeted.

We can add lot more features to add proper support like lipo and stuff but at least this enables you to try Mun on iDevices!

NOTE: I have *not* tested this on an actual iDevice so I don't know if this actually works on the real hardware. 
NOTE: As mentioned in https://github.com/mun-lang/mun/issues/417#issuecomment-1176735248 I don't think this will pass App Store rules because loading of dynamic libraries at runtime is not permitted.

Closes #417

### DIFF
--- a/crates/mun_codegen/src/apple.rs
+++ b/crates/mun_codegen/src/apple.rs
@@ -1,4 +1,6 @@
-use once_cell::sync::OnceCell;
+use once_cell::sync::Lazy;
+use std::collections::HashMap;
+use std::sync::Mutex;
 use std::{
     env, io,
     path::{Path, PathBuf},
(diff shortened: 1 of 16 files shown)


crates/mun_target/src/spec/x86_64_unknown_linux_gnu.rs
```diff
@@ -1,21 +1,15 @@
...
-        target_vendor: "unknown".to_string(),
+        pointer_width: 64,
         arch: "x86_64".to_string(),
         data_layout: "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
             .to_string(),
-        linker_flavor: LinkerFlavor::Ld,
-        options: base,
-    })
+        options: TargetOptions {
+            cpu: "x86_64".into(),
```
baszalmstra, line 11:
```suggestion
            cpu: "x86-64".into(),
```

crates/mun_codegen/src/apple.rs
```diff
@@ -1,4 +1,6 @@
-use once_cell::sync::OnceCell;
+use once_cell::sync::Lazy;
+use std::collections::HashMap;
+use std::sync::Mutex;
```
Wodann, line 3: Use `parking_lot::Mutex`. It's guaranteed to succeed

crates/mun_target/src/spec/apple_sdk_base.rs
```diff
@@ -0,0 +1,49 @@
+use crate::spec::TargetOptions;
+
+use Arch::*;
+
+#[allow(non_camel_case_types, dead_code)]
```
Wodann, line 5: Is this `dead_code`? It seems to be used underneath

Wodann (changes requested):
Some small comments. Otherwise LGTM :)

crates/mun_codegen/src/apple.rs
baszalmstra: Nice fixed!

crates/mun_target/src/spec.rs
```diff
@@ -18,33 +20,15 @@ pub struct Target {
...
-    /// The name of the environment
-    pub target_env: String,
+    pub pointer_width: u32,
 
-    /// The name of the vendor
-    pub target_vendor: String,
-
-    /// The name of the architecture. For example "x86" or "x86_64"
+    /// The name of the architecture. For example "x86" or "x86_64", "arm", "aarch64"
     pub arch: String,
```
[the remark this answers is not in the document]
baszalmstra: I ran it again and nothing changes, I also don't see strange formatting in the browser. What do you see?

crates/mun_target/src/spec/apple_sdk_base.rs
baszalmstra: The enum itself is used, but not all variants are constructed. I kept them in because these are all the supported architectures at the moment.

crates/mun_target/src/spec.rs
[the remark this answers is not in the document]
Wodann: It might have been the mobile app that was messing up then I was seeing some member variables without any indentation at all.

Wodann (approved):

PR #427, opened 2022-07-20
feat: add `Type` API to mitigate issues around memory

### LINKED ISSUE #423: Cyclic type reference causes hang
A runtime bug was introduced when loading an assembly that contains a cyclic datastructure. The following code is valid Mun code because `Foo` is garbage collected type (even though its impossible to create this type at runtime).

```rust
pub struct Foo {
    foo: Foo
}
```

When loading a munlib at runtime that contains this type the runtime hangs. This is due to the code in assembly.rs:

https://github.com/mun-lang/mun/blob/fa07c5c9ae85b27cff5faf7e873583bf10e2de86/crates/mun_runtime/src/assembly.rs#L215-L223

When loading the type `Foo`, if the type `Foo` is not yet in the `TypeTable` loading the above snippet will fail.

We could do something by first creating the `Arc<TypeInfo>` and then injecting it somehow but it raises another interesting question. By storing the `Arc<TypeInfo>` in the field, we create a cycle. However, storing the type of a field as a `Weak<TypeInfo>` is also not an option in a case like this:

```rust
pub struct Foo {
    bar: Bar
}

pub struct Bar {
    foo: Foo
}
```

If in the case above we only hold an `Arc<TypeInfo>` of `Bar`, the type info of `Foo` might be dropped which is not what we want.

I see two approaches in solving this:

* Always retain all the TypeInfos ever allocated. Memory will grow over time, but I doubt it will become a major issue.
* Garbage collect TypeInfos. 

The first option is definitely the easiest to implement, I recommend we first take that approach and optimize later. WDYT?

baszalmstra:
This PR refactors the Type system. It exposes only: `Type` which is now your one and only handle for type information. Internally it uses reference counting and a system to ensure that cross-referencing types are not deallocated. 

I also had to rewrite the FFI code for types because of this so I also made all the FFI code and tests a little more ergonomic/easy to use.

The only thing missing currently is the C++ code that accompanies the changes.

This does fix issue #423 .

Closes #423


crates/mun_runtime/src/function_info.rs
```diff
@@ -1,6 +1,8 @@
-use std::{ffi::c_void, sync::Arc};
+use std::ffi::c_void;
```
Wodann, line 1: Merge imports

crates/mun_runtime_capi/src/runtime.rs
```diff
@@ -1,20 +1,46 @@
...
-use memory::TypeInfo;
-use runtime::{FunctionDefinition, FunctionPrototype, FunctionSignature, Runtime};
-use std::{
-    ffi::{c_void, CStr},
-    os::raw::c_char,
-    sync::Arc,
-};
+use memory::Type as RustType;
+use runtime::{FunctionDefinition, FunctionPrototype, FunctionSignature};
+use std::mem::ManuallyDrop;
```
Wodann, line 9: Merge imports

Wodann (changes requested):

crates/mun_runtime/src/assembly.rs
```diff
@@ -129,7 +129,7 @@ impl Assembly {
                     })
                     .collect::<Result<Vec<_>, _>>()
                     .with_context(|| {
-                        format!("failed to link function '{}'", fn_prototype.name())
+                        format!("failed to link function '{}' b", fn_prototype.name())
```
Wodann, line 132: a, b, c markers are still there for the `failed to link function` occurrences.

Wodann (changes requested):

Wodann (approved):

PR #431, opened 2022-08-04
WIP: refactor: memory mapping

Wodann:
Creates a clearer separation between the calculation of a memory mapping and the execution thereof. Also, adds support for array memory mapping.

Supported array mappings are:
- [f64] <-> f64 (array from/to primitive)
- [Foo] <-> Foo (array from/to struct)


Wodann:
See: https://github.com/baszalmstra/mun/pull/20/files

PR #433, opened 2022-08-06
feat: use cargo-nextest in ci

baszalmstra:
Use [`cargo nextest`](https://nexte.st/) in CI


baszalmstra:
@Wodann I think the output of nextest is much nicer than regular cargo test for sure but I don't know if the costs outway of "something extra" and non-standardness outway the benefits. If you don't mind I would like to try for a while and revert if we don't feel this adds more than it costs. WDYT?

PR #444, opened 2022-08-27
feat: add M1 support

Wodann:


crates/mun_runtime/src/lib.rs
```diff
@@ -16,7 +16,7 @@ mod reflection;
 use anyhow::Result;
 use dispatch_table::DispatchTable;
 use garbage_collector::GarbageCollector;
-use log::{debug, error, info};
+use log::{debug, error, info, warn};
```
baszalmstra, line 19: `warn` is unused.

```suggestion
use log::{debug, error, info};
```

baszalmstra (approved):

baszalmstra:
Im looking into the disk space issue.

baszalmstra:
Our build is using more than 14gb. Do you think we can remove some dependencies? Or maybe we can not build the dependencies for the benchmarks? We are not running them anyway in CI and cargo check still checks them (I think).

Wodann:
> Our build is using more than 14gb. Do you think we can remove some dependencies? Or maybe we can not build the dependencies for the benchmarks? We are not running them anyway in CI and cargo check still checks them (I think).

I actually started looking through our dependencies with tools, in order to remove them. The tricky part is that our inkwell dependency prohibits us from using any of the normal tooling, as it cannot resolve its version naming.

PR #478, opened 2022-11-14
feat: use llvm-cov for coverage

baszalmstra:
Use llvm-cov instead of tarpauling to compute coverage. LLVM-cov uses recently stabilized code coverage metrics emitted by the rust compiler. This also then works on Windows and MacOS and can be collected while running the regular tests.

I did have to change some tests to no longer use the name of the thread that is executing the tests. This is because that behavior (which was unstable in the first place) doesn't work with llvm-cov.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -4,9 +4,9 @@ gen-syntax        = "run --package tools --bin tools -- gen-syntax"
 gen-runtime-capi  = "run --package tools --bin tools -- gen-runtime-capi"
 gen-abi           = "run --package tools --bin tools -- gen-abi"
 
-[profile.min-artifact-size-debug]
+[profile.llvm-cov-target]
 inherits = "dev"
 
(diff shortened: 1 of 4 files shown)


Wodann (commented):
I'm curious to see whether there is a performance overhead compared to `nextest`

baszalmstra:
Looks like its working properly! Except on windows. I figured that llvm-cov is using another cargo profile which I have to make sure is also used during doctesting, otherwise the build will take much longer and disc space will run out.

Wodann:
The dreaded out of storage issue is back 😢

baszalmstra:
> The dreaded out of storage issue is back 😢
> 
> Did you still have options to try to reduce storage usage? Can we split main tests vs example tests?

Yeah! I already have an idea as outlined in the comment above!

Wodann (approved):

PR #516, opened 2023-02-06
fix: install libpolly explicitly

baszalmstra:
Explicitly install `libpolly`


Wodann:
Do you want to preserve anything from this PR, after the fix in https://github.com/mun-lang/mun/pull/523 already having been incorporated?

baszalmstra:
Nope! Thanks for fixing!

PR #578, opened 2025-01-07
feat: assoc method completion

baszalmstra:
This PR refactors how we analyze the context of completions and adds completion of associated methods.

There is some unused code in this PR which I need for adding more completions (like methods and use statements). I didnt add them to this PR yet to make sure it doesnt grow to big.


crates/mun_language_server/src/completion/unqualified_path.rs
```diff
@@ -10,6 +10,8 @@ use super::{CompletionContext, Completions};
 /// }
 /// ```
 pub(super) fn complete_unqualified_path(result: &mut Completions, ctx: &CompletionContext<'_>) {
+    dbg!(ctx.is_trivial_path, ctx.is_param, ctx.is_path_type);
```
Wodann, line 13: Is this intended to be left?

Wodann (approved):
Overall LGTM. One question

crates/mun_language_server/src/completion/unqualified_path.rs
baszalmstra: Woops! Nope!

```suggestion
```

PR #580, opened 2025-01-13
refactor: use inline completion snapshots

baszalmstra:
Adds completion relevance system and complete locals somewhat higher than non-locals.

Also contains some general refactors I stumbled upon along the way.

### DIFF
--- a/crates/mun_language_server/src/completion/expr.rs
+++ b/crates/mun_language_server/src/completion/expr.rs
@@ -46,7 +46,7 @@ pub(super) fn complete_expr_path(
 
 #[cfg(test)]
 mod tests {
-    use crate::completion::{test_utils::completion_string, CompletionKind};
+    use crate::completion::{test_utils::completion_string};
 
     #[test]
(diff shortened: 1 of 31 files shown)


crates/mun_language_server/src/completion/expr.rs
```diff
@@ -57,9 +57,12 @@ mod tests {
             let foo_bar = 0;
             f$0
         }
-        "#,
-            Some(CompletionKind::Reference)
-        ));
+        "#
+        ), @r###"
```
Wodann, line 61: What does the `@` symbol mean?

Wodann (approved):
One question. Otherwise LGTM!

crates/mun_language_server/src/completion/expr.rs
baszalmstra: its a marker for insta so it knows where to place the inline snapshot.

Wodann: Thanks!

PR #581, opened 2025-02-09
build: improve devcontainer setup

Wodann:
This PR also solves a warning about the inclusion of lints in the config file


Wodann:
The failing CI job is also failing on `main`

